PR #23, opened 2024-05-02
Adam dev

Asanto32:
This pull request addresses Task 1 and Task 2: implementing new models classes, and readers functionality.   It contains unit tests for these modules (including sample data). There has been some file pruning, as well as updates to the poetry/pyproject requirements.

PR #24, opened 2024-05-03
Implements eucleidean_norm metric calculation

### LINKED ISSUE #22: Task: Implement euclidean_norm in metrics.py
### Description

The goal is to implement a function that will calculate the value of ENMO (Euclidean Norm Minus One) from 3 dimensional accelerometer data. As the name suggests the metric is the Euclidean norm of the 3D acceleration signal minus 1g(1 standard gravity). As a measurement of physical activity, ENMO is one of the key metrics calculated by wristpy and ggir and is of high scientific interest in the fields of actigraphy and sleep research. 

The input of the function will be a Measurement object containing the acceleration data(an np array) and timestamps(a polars series) the function will output another Measurement object where the measurement will be the ENMO values at corresponding timestamps. 

### Tasks

- [x] The function should take as input an instance of the Measurement class, where the measurement attribute is acceleration in the form of an np.array and the time attribute is a polars Series of datetime.datetime objects. 
- [x] the function should output a Measurement object where the measurement attribute is enmo and the time attribute is the unmodified polars Series derived from acceleration.time. 

### Freeform Notes

_No response_

frey-perez:
- Resolves #22 

This PR implements the changes in Task #22 Implement eucleidean_norm in metrics.py. It creates the metrics.py module, and it's first function euclidean_norm, in addition to the associated unit tests. Dependencies were updated as needed.


src/wristpy/processing/metrics.py
```diff
@@ -0,0 +1,27 @@
...
+        datetime.datetime objects.
+
+    Returns:
+        A Measurement object containing the calculated ENMO values and the 
+        associated time stamps taken from the input.
+    """
+    #compute euclidean norm - 1
+    enmo = np.linalg.norm(acceleration.measurements, axis = 1) - 1
+
+    #GGIR replaces negative values with 0
```
ReinderVosDeWael, line 24: Try to avoid in-line comments; no need to say what the code does when the code already speaks for itself :). In-line comments should be used for the rare exception that there's no way to write the code clearly.

That being said, decisions like removing negative numbers do warrant some explanation. This would generally go in the docstring though i.e. between the top-level summary sentence and the Args block e.g. something like:

```python
def euclidean_norm(acceleration: Measurement)-> Measurement:
    """Compute ENMO, the Euclidean Norm Minus One (1 standard gravity unit).

    Negative ENMO values are set to zero because [add reason here; "GGIR does it" is not a reason ;) ] 

    Args:
        acceleration: the three dimensional accelerometer data. A Measurement object,
        it will have two attributes. 1) measurements, containing the three dimensional 
        accelerometer data in an np.array and 2) time, a pl.Series containing 
        datetime.datetime objects.
    ...
    """
 ```

File: tests/unit/test_metrics.py [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+            (np.ones(TEST_LENGTH)*3, np.zeros(TEST_LENGTH), np.zeros(TEST_LENGTH), 2)
+        ]
+)
+def test_euclidean_norm(
+        x: np.ndarray, y: np.ndarray, z: np.ndarray, expected_enmo: float
+        )->None:
+    """Tests the euclidean norm function."""
+    dummy_date = datetime(2024, 5, 2)
+    dummy_datetime = pl.Series("time", [dummy_date] * TEST_LENGTH)
+    #assemble Measurement object with test array and dummy datetime
```
ReinderVosDeWael, line 28: No need for inline-comment.

src/wristpy/processing/metrics.py
```diff
@@ -0,0 +1,27 @@
+"""Calculate base metrics, anglez and enmo."""
+
+import numpy as np
+
+from wristpy.core.models import Measurement
+
+
+def euclidean_norm(acceleration: Measurement)-> Measurement:
```
ReinderVosDeWael, line 8: I know you copied this from the diagram; so this is my bad, but since this isn't strictly a euclidean norm I wouldn't use this name. 

How about we name this `def euclidean_norm_minus_one` or `def euclidean_norm_min_one` if you want to shorten it a bit. Feel free to use anything else as long as the functionality is self-evident from the name.

tests/unit/test_metrics.py
```diff
@@ -0,0 +1,39 @@
...
+
+TEST_LENGTH = 100
+
+@pytest.mark.parametrize(
+        "x,y,z, expected_enmo",
+        [
+            (np.zeros(TEST_LENGTH), np.zeros(TEST_LENGTH), np.zeros(TEST_LENGTH), 0),
+            (np.ones(TEST_LENGTH), np.zeros(TEST_LENGTH), np.zeros(TEST_LENGTH), 0),
+            (np.ones(TEST_LENGTH)*2, np.zeros(TEST_LENGTH), np.zeros(TEST_LENGTH), 1),
+            (np.ones(TEST_LENGTH)*3, np.zeros(TEST_LENGTH), np.zeros(TEST_LENGTH), 2)
```
ReinderVosDeWael, line 19: Love the parameterization; though try to limit it to unique cases. For example, I see three unique cases here that could be tested: 1) euclidean norm-1 is negative, so result is 0. 2) euclidean norm-1 is zero, result should still be 0, and 3) euclidean norm is positive. In at least one of these cases (probably the latter), try to use at least one test that has non-zero input for all three.

tests/unit/test_metrics.py
```diff
@@ -0,0 +1,39 @@
...
+    #assemble Measurement object with test array and dummy datetime
+    test_acceleration = Measurement(
+        measurements= np.column_stack((x, y, z)),
+        time = dummy_datetime)
+
+    enmo_results = euclidean_norm(test_acceleration)
+
+    
+
+    assert np.all( np.isclose(
```
ReinderVosDeWael, line 37: Assert that `input_time == output_time` and that the shape of `measurements` is as expected, should be `(TEST_LENGTH)` I think?.

ReinderVosDeWael (changes requested):
Looks great for your first PR :). 

Do you use VSCode? If yes, try going into your settings and setting it to auto-format on saving a file, and set your formatter to ruff. I'm not sure why CI Ruff isn't failing on this (@nx10?) but it should fail on `return Measurement(measurements= enmo, time = acceleration.time)`

For simplicity on this PR, you can try running `ruff format [FILENAME]` to correct the formatting on the Python files. 

Also here's a question I'd like you to think about (though don't make changes based on this (yet)): `euclidean_norm`, as implemented now, would accept measurements of any shape provided that there are at least two dimensions. Is that OK? Is it a problem? Why? Should this behavior be reflected in the docstring, or should the docstring reflect the primary usage within the application (as it does now)? 

**Note for merging:** this PR depends on #23; hold merging until that PR is merged.

tests/unit/test_metrics.py
frey-perez: Okay i'll add a case (should i add or just change the third case?) where I'll use a pythagorean quadruple 1,2,2,3. That does make me wonder, although this is still pretty easy to verify, do I have to at some point "show my work" when coming up with ground truth values that are based off of calculations like this ?

ReinderVosDeWael: I wouldn't expect you to show this in the documentation (I'd expect a reader of the code to be able to look up a euclidean norm algorithm themselves), that being said, you need to be able to verify the correctness of any algorithm that you implement. This can be done by hand (relatively straightforward here, see below for [1,2,2,3]), or for more complex algorithms by verifying against known cases and/or other implementations. 
 
```
sqrt(1^2 + 2^2 + 2^2 + 3^2) - 1
= sqrt(1+4+4+9) - 1
= sqrt(18) - 1
= sqrt(9) * sqrt(2) - 1
= 3 * sqrt(2) - 1
```

FYI [1,2,2,4] makes the calculation a lot easier :).

src/wristpy/processing/metrics.py
```diff
@@ -0,0 +1,30 @@
...
+from wristpy.core.models import Measurement
+
+
+def euclidean_norm_min_one(acceleration: Measurement) -> Measurement:
+    """Compute ENMO, the Euclidean Norm Minus One (1 standard gravity unit).
+
+    Negative values of ENMO are set to zero because ENMO is meant as a measure of
+    physival activity. Negative values steming from low levels of activity or noise from
+    the device would have no meaningful interpretation in this context and would be
+    detrimental to the intended analysis.
```
Asanto32, line 14: I might change this description of the wrapping to zero slightly. It is more due to imperfect calibration + device noise. In theory, if the sensor was perfectly calibrated ENMO should never be negative (unless you were in a zero-_g_ room or in flight?). Not necessarily related to low levels of physical activity.

src/wristpy/processing/metrics.py
```diff
@@ -0,0 +1,30 @@
+"""Calculate base metrics, anglez and enmo."""
+
+import numpy as np
+
+from wristpy.core.models import Measurement
+
+
+def euclidean_norm_min_one(acceleration: Measurement) -> Measurement:
```
Asanto32, line 8: Why not just name the function euclidean_norm_minus_one? It only saves two characters?

tests/unit/test_metrics.py
```diff
@@ -0,0 +1,46 @@
...
+        (np.ones(TEST_LENGTH), np.ones(TEST_LENGTH) * 2, np.ones(TEST_LENGTH) * 2, 2),
+    ],
+)
+def test_euclidean_norm_min_one(
+    x: np.ndarray, y: np.ndarray, z: np.ndarray, expected_enmo: float
+) -> None:
+    """Tests the euclidean norm function."""
+    dummy_date = datetime(2024, 5, 2)
+    dummy_datetime = pl.Series("time", [dummy_date] * TEST_LENGTH)
+    test_acceleration = Measurement(
```
Asanto32, line 30: in theory you can use my readers.unix_epoch_time_to_polars_datetime function here, probably not necessary

tests/unit/test_metrics.py
```diff
@@ -0,0 +1,46 @@
...
+        measurements=np.column_stack((x, y, z)), time=dummy_datetime
```
Asanto32, line 31: Actually, this is interesting that Measurement accepts this. All the timestamps are the exact same value. We validate that the time  field is sorted, and apparently this counts. @ReinderVosDeWael should we concerned about this in our Measurement model?

tests/unit/test_metrics.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+
+    enmo_results = euclidean_norm_min_one(test_acceleration)
+
+    assert np.all(
+        np.isclose(enmo_results.measurements, expected_enmo)
+    ), f"Expected {expected_enmo}"
```
Asanto32, line 38: I know I initially used `np.isclose` but for these simple cases should we use `np.array_equal`?

Asanto32 (commented):
I didn't approve it just yet @ReinderVosDeWael but I have no real concerns.

tests/unit/test_metrics.py
nx10: Never use `array_equal` (or `==`) on floating point values.

```Python
import numpy as np

arr1 = np.array([0.1]) + np.array([0.2])
arr2 = np.array([0.3])

print("arr1", arr1)
print("arr2", arr2)
print("equal", np.equal(arr1, arr2))
print("array_equal", np.array_equal(arr1, arr2))
print("isclose", np.isclose(arr1, arr2))
print("allclose", np.allclose(arr1, arr2))
```
```
arr1 [0.3]
arr2 [0.3]
equal [False]
array_equal False
isclose [ True]
allclose True
```

src/wristpy/processing/metrics.py
ReinderVosDeWael: Think that might've been my suggestion in case Freymon felt the name was getting _too_ long. Though that may have been my Dutch slipping through (Dutch word for minus is min), forgetting that min is more commonly used to abbreviate minimum. So yeah, I disagree with past Reinder :P

tests/unit/test_metrics.py
ReinderVosDeWael: Depends; is there ever a case where you expect the watch to return multiple measurements for the same timepoint? If no, then yeah I'd double check that all time values are unique.

tests/unit/test_metrics.py
ReinderVosDeWael: I prefer this actually - reduces the number of dependencies.

tests/unit/test_metrics.py
Asanto32: Definitely should not have the exact same timestamp for multiple data points.

Yes. I can create a new task to add that as part of the time field validator. That being said it will then cause this unit test to fail  on the Measurement() initialization?

ReinderVosDeWael (approved):
LGTM - after you fix your import error ;)

tests/unit/test_metrics.py
ReinderVosDeWael: That's fine - since we're aware of the issue already you could get ahead of it now @frey-perez if you want, but in general old tests failing due to changes in architecture is fine, so long as the tests are updated to reflect that new architecture.  

As far as I'm concerned it can be handled in either PR.

PR #25, opened 2024-05-06
Fix ruff test not detecting bad formatting

ReinderVosDeWael:
The Ruff checker is missing formatting issues at the moment; this PR addresses this bug without altering the newly introduced error, as this should be handled by #23. The unit test failure exists in the development branch and should be considered non-blocking for merging.


Asanto32 (approved):
I assume this is ok to approve

PR #31, opened 2024-05-07
Implementing moving_mean function in metrics.py

Asanto32:
This PR implements the moving_mean function in metrics.py as covered in #26


ReinderVosDeWael (changes requested):
Can elaborate on these comments tomorrow; stuck in meetings the rest of the day. But figured I'd get you the review already.

src/wristpy/processing/metrics.py
```diff
@@ -0,0 +1,59 @@
...
+    The input must be a Measurement instance. The window size must be a posititve
+    non-zero integer. The function converts the epoch_length to a string that can
+    be used by polars group_by_dynamic. We then create a polars dataframe that is
+    a concatenation of the sensor data and the equivalent time data. We group
+    the data by creating a windowbased on the timestamp using the "every" function
+    and then take the mean of that group.
+
+    Args:
+        array: The Measurement object with the sensor data we want to take the mean of
+        epoch_length: The length of the window to calculate the moving mean over.
```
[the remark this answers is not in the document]
Asanto32: Discussed in meeting this morning

src/wristpy/processing/metrics.py
```diff
@@ -0,0 +1,59 @@
...
+    if epoch_length < 1:
+        raise ValueError("Epoch length must be greater than 0")
+
+    window_size_seconds = str(epoch_length) + "s"
+
+    measurement_polars_df = pl.concat(
+        [pl.DataFrame(array.measurements), pl.DataFrame(array.time)], how="horizontal"
+    )
+
+    measurement_polars_df = measurement_polars_df.with_columns(
```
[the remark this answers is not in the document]
Asanto32: Per the documentation of pl.expr.set_sorted(): 
Flags the expression as ‘sorted’.
Enables downstream code to user fast paths for sorted arrays.

I believe it is required explicitly to use group_by_dynamic. At least in all my previous testing it was needed.

src/wristpy/processing/metrics.py
```diff
@@ -0,0 +1,59 @@
...
+        pl.all().exclude(["time"]).drop_nans().mean(),
+    ]
+
+    measurement_df_mean = measurement_polars_df.group_by_dynamic(
+        index_column="time", every=window_size_seconds
+    ).agg(take_mean_expression)
+
+    return Measurement(
+        measurements=measurement_df_mean.drop("time").to_numpy(),
+        time=measurement_df_mean["time"],
```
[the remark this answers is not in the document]
Asanto32: No, time does change. The returned Measurement.time will only have time stamps every five seconds, ex. of 00:00, 00:05, 00:10....

The returned Measurement from moving_mean is downsampled so it is  shorter by a factor of ~(sampling_rate *epoch_length)

tests/test_metrics.py
```diff
@@ -0,0 +1,74 @@
...
+    test_data = np.arange(0, signal_length)
+    test_time = readers.unix_epoch_time_to_polars_datetime(
+        np.arange(0, signal_length), "s"
+    )
+
+    test_measurement = Measurement(measurements=test_data, time=test_time)
+    test_measurement_mean = metrics.moving_mean(test_measurement, epoch_length=5)
+
+    assert np.array_equal(
+        test_measurement_mean.measurements.flatten(), np.array([2.0, 7.0, 12.0, 17.0])
```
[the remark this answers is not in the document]
Asanto32: The issue is pl.to_numpy() returns 2D array (even if input was 1D), so the shape returned is (4x1). I have changed expected to be (4,1) as well. I have also added a test to ensure 3D array maintains it's shape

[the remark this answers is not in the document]
ReinderVosDeWael: Why not just reshape to the expected output within the function i.e. `pl.to_numpy().reshape(input.shape)? Or is there a reason why this should always become 2D even if the input is 1D?

src/wristpy/core/computations.py
```diff
@@ -0,0 +1,51 @@
...
+        array: The Measurement object with the sensor data we want to take the mean of
+        epoch_length: The length, in seconds, of the window.
+
+    Returns:
+        The moving mean of the array in a new Measurement instance.
+
+    Raises:
+        ValueError: If the epoch length is not an integer or is less than 1.
+    """
+    if not isinstance(epoch_length, int):
```
frey-perez, line 21: not sure if we need the isinstance check, we have the type hinting and this is the kind of function only the developer would use directly right?

tests/test_computations.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+    expected_mean = np.array(
+        ([[6.0, 7.0, 8.0], [21.0, 22.0, 23.0], [36.0, 37.0, 38.0], [51.0, 52.0, 53.0]])
+    )
+    test_measurement = Measurement(measurements=test_data, time=test_time)
+
+    test_measurement_mean = computations.moving_mean(test_measurement, epoch_length=5)
+
+    assert np.array_equal(test_measurement_mean.measurements, expected_mean)
+    assert test_measurement_mean.measurements.shape[1] == test_data.shape[1]
```
frey-perez, line 61: Might be worth adding tests to make sure time is the appropriate shape as well. (in both 1D and 3D versions).

tests/test_computations.py
```diff
@@ -0,0 +1,61 @@
...
+    test_data = np.arange(0, signal_length)
+    test_time = readers.unix_epoch_time_to_polars_datetime(
+        np.arange(0, signal_length), "s"
+    )
+    expected_mean = np.array([2.0, 7.0, 12.0, 17.0])
+    test_measurement = Measurement(measurements=test_data, time=test_time)
+
+    test_measurement_mean = computations.moving_mean(test_measurement, epoch_length=5)
+
+    assert np.array_equal(test_measurement_mean.measurements, expected_mean)
```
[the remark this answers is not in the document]
Asanto32: Strange, I thought I changed this before you reviewed it...And it even says Outdated next to the comment. But yes, resolved.

src/wristpy/core/computations.py
```diff
@@ -0,0 +1,51 @@
+"""Calculate base metrics, enmo, anglez, and non-wear detection."""
```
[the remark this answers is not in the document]
Asanto32: Yes, good catch.

src/wristpy/core/computations.py
Asanto32: I can see both ways since a non-int would cause it to crash. I was also debating just doing `int(epoch_length)` any non-int instead of raising an error. Thoughts @ReinderVosDeWael ?

ReinderVosDeWael: People who can't read type hints aren't our problem. If the type hint says it is an int, it's an int.

ReinderVosDeWael: Less tongue in cheek, yeah these kinda checks really should be reserved for places where a non technical user enters the program e.g. where they supply us with a file.

tests/test_computations.py
Asanto32: Hmmm... the length of the series changes, but I can hardcode to check that it is the expected length relative to the input.

src/wristpy/core/computations.py
Asanto32: Got it. Will make the changes and then I think everything should be good

tests/test_computations.py
```diff
@@ -0,0 +1,59 @@
+"""Test the function of computations module."""
+
+import numpy as np
+import pytest
+
+from wristpy.core import computations
```
ReinderVosDeWael, line 6: Minor point: Any reason why you import some things by module (e.g. computations) and some by class/function (e.g. Measurement)?

I'd keep this consistent across your code; common in most Python open source is importing the class/function, my preference follows the [google python style guide](https://google.github.io/styleguide/pyguide.html#22-imports) where you only import modules (with a few exceptions). 

Pick whichever you prefer, but I'd pick one and make sure you and @frey-perez are consistent on this.

ReinderVosDeWael (approved):
LGTM - one very minor comment on styling.

tests/test_computations.py
ReinderVosDeWael: Assuming this isn't currently consistent across the codebase, an easy transition method is that if your PR affects an old file, you're responsible for correcting this in _that_ file.

PR #34, opened 2024-05-10
Adding moving_std to computations

### LINKED ISSUE #28: Task: Implement moving_std function in core/computations.py
### Description

The goal is to implement a function that will calculate the moving  standard deviation of a measurement object and return a new, downsampled, measurement object. This moving standard deviation (std) will be implemented using a dynamic time window; we will use the exact time stamps and not the number of samples to find the window size corresponding to the desired window length, in seconds. The moving_std function is necessary as part of the calibration and non-wear detection algorithms.

The inputs of the function will be a Measurement object, containing the specific sensor data that we want to downsample, the epoch_length (int) that provides the number of seconds that defines the window size. The function returns a new Measurement object with the downsampled sensor data and the new, corresponding time information.

### Tasks

- [x] The function will take as input an instance of the Measurement class and an integer defining the window length, in seconds.
- [x] The function will implement a dynamically sized window based on the timestamps in Measurement.time and then take the std of each window.
- [x] The result of the moving std will be a new instance of the Measurement class and this will be returned.

### Freeform Notes

_No response_

Asanto32:
- resolves #28

Implement the moving standard deviation function in computations.py. Code for the function and tests are based on the moving_mean function. 
Minor fixes to test_computations: using global vars for SIGNAL_LENGTH and EPOCH_LENGTH


tests/unit/test_computations.py
```diff
@@ -0,0 +1,112 @@
...
+    assert np.isclose(
+        test_measurement_mean.time.shape[0], (test_time.shape[0] / EPOCH_LENGTH)
+    )
+
+
+def test_moving_std_epoch_length_is_negative() -> None:
+    """Test error if the epoch length is negative."""
+    tmp_data = np.array([1, 2, 3])
+    time = readers.unix_epoch_time_to_polars_datetime(np.array([1, 2, 3]), "s")
+    tmp_Measurement = models.Measurement(measurements=tmp_data, time=time)
```
frey-perez, line 69: same comment as with the above test_moving_mean_epoch_length_is_negative function, I think this could just be tmp_measurement

tests/unit/test_computations.py
```diff
@@ -0,0 +1,112 @@
...
+        computations.moving_std(tmp_Measurement, epoch_length=-EPOCH_LENGTH)
+
+
+def test_moving_std_one_column() -> None:
+    """Test the functionality of the moving std function for 1D Measurement."""
+    test_data = np.ones(SIGNAL_LENGTH)
+    test_time = readers.unix_epoch_time_to_polars_datetime(
+        np.arange(0, SIGNAL_LENGTH), "s"
+    )
+    expected_std = np.array([0.0, 0.0, 0.0, 0.0])
```
frey-perez, line 81: This might be superfluous, but might be worth trying parameters that give a non-zero std.

frey-perez (commented):
Looks good to me @ReinderVosDeWael, only noticed a couple of minor docstring/ formatting things, and a suggestion for some extra tests if necessary.

tests/unit/test_computations.py
```diff
@@ -0,0 +1,112 @@
...
+
+SIGNAL_LENGTH = 20
+EPOCH_LENGTH = 5
+
+
+def test_moving_mean_epoch_length_is_negative() -> None:
+    """Test error if the epoch length is negative."""
+    tmp_data = np.array([1, 2, 3])
+    time = readers.unix_epoch_time_to_polars_datetime(np.array([1, 2, 3]), "s")
+    tmp_Measurement = models.Measurement(measurements=tmp_data, time=time)
```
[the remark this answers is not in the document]
ReinderVosDeWael: Yup, c.f. https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations, its a local variable.

tests/unit/test_computations.py
```diff
@@ -0,0 +1,112 @@
...
+
+    with pytest.raises(ValueError):
+        computations.moving_mean(tmp_Measurement, epoch_length=-EPOCH_LENGTH)
```
ReinderVosDeWael, line 20: If all that's relevant here is that EPOCH_LENGTH is a negative number, just write -1 or some such; in this set-up I'd have to scroll up to figure out what this is.

tests/unit/test_computations.py
```diff
@@ -0,0 +1,112 @@
...
+    test_measurement = models.Measurement(measurements=test_data, time=test_time)
+
+    test_measurement_mean = computations.moving_mean(
+        test_measurement, epoch_length=EPOCH_LENGTH
+    )
+
+    assert np.allclose(test_measurement_mean.measurements, expected_mean)
+    assert test_measurement_mean.measurements.ndim == expected_mean.ndim
+    assert np.isclose(
+        test_measurement_mean.time.shape[0], (test_time.shape[0] / EPOCH_LENGTH)
```
ReinderVosDeWael, line 39: You could extract this into a variable inside the arrange e.g. `expected_time_shape = ...` so the assertion reads more simply.

Also, the shape array should contain integers, not floats, so use np.isequal in this case. Make sure to set expected_time_shape to an int (and to floor/ceil correctly) before asserting.

tests/unit/test_computations.py
```diff
@@ -0,0 +1,112 @@
...
+    test_measurement = models.Measurement(measurements=test_data, time=test_time)
+
+    test_measurement_mean = computations.moving_mean(
+        test_measurement, epoch_length=EPOCH_LENGTH
+    )
+
+    assert np.allclose(test_measurement_mean.measurements, expected_mean)
+    assert test_measurement_mean.measurements.shape[1] == test_data.shape[1]
+    assert np.isclose(
+        test_measurement_mean.time.shape[0], (test_time.shape[0] / EPOCH_LENGTH)
```
ReinderVosDeWael, line 61: See the isclose and int comment above.

tests/unit/test_computations.py
```diff
@@ -0,0 +1,112 @@
...
+    )
+    test_measurement = models.Measurement(measurements=test_data, time=test_time)
+
+    test_measurement_mean = computations.moving_mean(
+        test_measurement, epoch_length=EPOCH_LENGTH
+    )
+
+    assert np.allclose(test_measurement_mean.measurements, expected_mean)
+    assert test_measurement_mean.measurements.shape[1] == test_data.shape[1]
+    assert np.isclose(
```
ReinderVosDeWael, line 60: See shape comment above.

tests/unit/test_computations.py
```diff
@@ -0,0 +1,112 @@
...
+    expected_std = np.array([0.0, 0.0, 0.0, 0.0])
+    test_measurement = models.Measurement(measurements=test_data, time=test_time)
+
+    test_measurement_std = computations.moving_std(
+        test_measurement, epoch_length=EPOCH_LENGTH
+    )
+
+    assert np.allclose(test_measurement_std.measurements, expected_std)
+    assert test_measurement_std.measurements.ndim == expected_std.ndim
+    assert np.isclose(
```
ReinderVosDeWael, line 90: See shape comment above.

tests/unit/test_computations.py
```diff
@@ -0,0 +1,112 @@
...
+    expected_std = np.zeros((4, 3))
+    test_measurement = models.Measurement(measurements=test_data, time=test_time)
+
+    test_measurement_std = computations.moving_std(
+        test_measurement, epoch_length=EPOCH_LENGTH
+    )
+
+    assert np.allclose(test_measurement_std.measurements, expected_std)
+    assert test_measurement_std.measurements.shape[1] == test_data.shape[1]
+    assert np.isclose(
```
ReinderVosDeWael, line 110: See comment above.

tests/unit/test_computations.py
ReinderVosDeWael: Definitely a good idea

tests/unit/test_computations.py
```diff
@@ -0,0 +1,112 @@
...
+    assert test_measurement_std.measurements.ndim == expected_std.ndim
+    assert np.isclose(
+        test_measurement_std.time.shape[0], (test_time.shape[0] / EPOCH_LENGTH)
+    )
+
+
+def test_moving_std_three_columns() -> None:
+    """Test the functionality of the moving std function for three column array."""
+    test_data = np.ones((SIGNAL_LENGTH, 3))
+    test_time = readers.unix_epoch_time_to_polars_datetime(
```
ReinderVosDeWael, line 98: Does this test need to depend on our reader functionality? Could you just create a time with datetime.datetime? Also applies to other occurrences of same code.

ReinderVosDeWael (changes requested):
Couple of minor comments; and +1 to @frey-perez 's comment on having the happy-path test use non-zero outcomes.

Asanto32 (commented):
FYI I had trouble seeing Reinder's initial comments while doing this review so maybe just cross-reference them to be sure everything has been covered from that initial review?

ReinderVosDeWael (approved):

PR #38, opened 2024-05-15
Implementing rolling median function for metrics.py

### LINKED ISSUE #30: Task: Implement rolling median function in metrics.py
### Description

The goal is to write a function that will calculate a rolling median on our acceleration data. Calculating a rolling median is an important data processing step. The rolling median helps to stabilize the raw actigraphy data and limit the distorting effects of outliers. The improved quality of the data will then benefit our angle-z calculations, as they will be based on more representative data. 

### Tasks

- [x] The function will take an instance of the Measurement class, containing the acceleration data in the form of a numpy 2D array, with sample number of rows and 3 (xyz)columns. The time attribute will be a polars Series of datetime.datetime. objects.
- [x] The function will take as an optional input an int for the window_size. The default value will be 51. 
  - [ ] The sliding window must be centered, meaning logic to handle edge cases where the window will partly extend beyond the data range will be required. We will use a nan-mean i.e. the window gets smaller at the edges. 
- [x] The  function will output an instance of the Measurement class containing the rolling_median_acceleration data, and the unmodified time data from the input. 
  - [ ] The shape of rolling_median_acceleration output and the acceleration data input should be the same.

### Freeform Notes

_No response_

frey-perez:
- resolves #30 

This function implements the rolling median function, a function meant to pre-process acceleration data before determining angle_relative_to_horizontal values. All relevant unit tests are included.


nx10:
Why not use https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.rolling_median.html and other polars.rolling_* methods instead of implementing all of these?

- https://github.com/childmindresearch/wristpy/pull/31
- https://github.com/childmindresearch/wristpy/pull/34
- https://github.com/childmindresearch/wristpy/pull/38

src/wristpy/processing/metrics.py
```diff
@@ -53,3 +56,59 @@ def angle_relative_to_horizontal(
...
+
+    if window_size % 2 == 0:
+        window_size += 1
+        warnings.warn(
+            f"window size must be odd, adjusting window size to be {window_size}",
+            UserWarning,
+        )
+
+    radius = window_size // 2
+    raw_data = acceleration.measurements
```
ReinderVosDeWael, line 101: Don't alias it.

tests/unit/test_metrics.py
```diff
@@ -79,3 +79,57 @@ def test_angle_relative_to_horizontal(
     assert angle_z_results.measurements.shape == (1,), (
         f"Expected anglez shape: {(1,)}," f"got({angle_z_results.measurements.shape})"
     )
+
+
+def test_rolling_median_wrong_ndims() -> None:
```
ReinderVosDeWael, line 84: See comment above; this should work.

tests/unit/test_metrics.py
```diff
@@ -79,3 +79,57 @@ def test_angle_relative_to_horizontal(
...
+    dummy_date = datetime(2024, 5, 2)
+    dummy_datetime = pl.Series("time", [dummy_date])
+    bad_array = np.array([1, 2, 3])
+    test_measurement = models.Measurement(measurements=bad_array, time=dummy_datetime)
+
+    with pytest.raises(ValueError):
+        metrics.rolling_median(test_measurement)
+
+
+def test_rolling_median_window_size() -> None:
```
ReinderVosDeWael, line 95: Let numpy handle this; we dont need to handle it.

src/wristpy/processing/metrics.py
```diff
@@ -53,3 +56,59 @@ def angle_relative_to_horizontal(
...
+        measurements data will retain it's shape, and the time data will be returned
+        unaltered.
+    """
+    if acceleration.measurements.ndim <= 1:
+        raise ValueError("the input arrary must be more than 1 dimension")
+
+    if window_size <= 1:
+        raise ValueError("window size must be greater than 1.")
+
+    if window_size % 2 == 0:
```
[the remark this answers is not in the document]
ReinderVosDeWael: If 1 add test, if 2 is covered by regular tests.

ReinderVosDeWael (changes requested):

ReinderVosDeWael:
> Why not use https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.rolling_median.html and other polars.rolling_* methods instead of implementing all of these?
> 
> * [Implementing moving_mean function in metrics.py #31](https://github.com/childmindresearch/wristpy/pull/31)
> * [Adding moving_std to computations #34](https://github.com/childmindresearch/wristpy/pull/34)
> * [Implementing rolling median function for metrics.py #38](https://github.com/childmindresearch/wristpy/pull/38)

Fair point - for now, as there's some design questions to be answered regarding handling of even window size and unit of window size, lets leave this for now. I'll make an issue out of it though so we can reassess this when those issues are resolved.

src/wristpy/core/computations.py
```diff
@@ -43,7 +43,54 @@ def moving_mean(array: Measurement, epoch_length: int = 5) -> Measurement:
...
+) -> models.Measurement:
+    """Applies moving median to acceleration data.
+
+    Args:
+        acceleration: the three dimensional accelerometer data. A Measurement object,
+        it will have two attributes. 1) measurements, containing the three dimensional
+        accelerometer data in an np.array and 2) time, a pl.Series containing
+        datetime.datetime objects.
+
+        window_size: The size of the overlapping window within which the median will be
```
ReinderVosDeWael, line 63: overlapping with what? Isn't it just the size of the window?

src/wristpy/core/computations.py
```diff
@@ -43,7 +43,54 @@ def moving_mean(array: Measurement, epoch_length: int = 5) -> Measurement:
...
+        unaltered.
+    """
+    measurements_polars_df = pl.concat(
+        [pl.DataFrame(acceleration.measurements), pl.DataFrame(acceleration.time)],
+        how="horizontal",
+    )
+    measurements_polars_df = measurements_polars_df.set_sorted("time")
+    offset = -((window_size // 2) + 1)
+    offset_str = str(offset) + "s"
+    moving_median_df = measurements_polars_df.select(
```
ReinderVosDeWael, line 79: This function will not be able to take the moving median of anything but acceleration as it will fail on data that isn't N-by-3. Are we sure that we never need this for any other measurement? If no, can we just copy the implementation of `moving_mean` as I believe this one is valid for any N-by-M array.

ReinderVosDeWael (changes requested):
Two relatively small comments, but see if you can make sure the `moving_median` matches the input implementation of `moving_mean` so we can use them interchangeably.

src/wristpy/core/computations.py
frey-perez: That's true I think this is easy to fix by using "*" instead of naming the cols. Then it will work on any amount of columns that comes with a timestamp col. 



here's the revised portion: 

 moving_median_df = measurements_polars_df.select(
        [
            pl.median("*").rolling(
                index_column="time", period=f"{window_size}s", offset=offset_str
            )
        ]
    )

    return models.Measurement(
        measurements=moving_median_df.drop("time").to_numpy(),
        time=measurements_polars_df["time"],
    )

frey-perez: That's true I can edit this to cover any amount of cols like so: 

 ```
moving_median_df = measurements_polars_df.select(
        [
            pl.median("*").rolling(
                index_column="time", period=f"{window_size}s", offset=offset_str
            )
        ]
    )

    return models.Measurement(
        measurements=moving_median_df.drop("time").to_numpy(),
        time=measurements_polars_df["time"],
    )
    
    
```

src/wristpy/core/computations.py
frey-perez: yes exactly, I'll make that edit.

ReinderVosDeWael (approved):

PR #46, opened 2024-05-31
Refactor readers.py for actfast 1.0

nx10:
Search sensor tables for individual sensors. This allows wristpy to work with any actfast supported watch/format.


nx10:
You could probably also delete a bunch of these tests now, but I'll leave those decisions to you.

PR #49, opened 2024-06-12
feat: Implementing 2023 GGIR non-wear detection algorithm

Asanto32:
Adding the `detect_nonwear` function to metrics.py. This function implements the 2023 GGIR non-wear detection algorithm to determine when the non-wear flag should be set to true (1) or false (0).


src/wristpy/processing/metrics.py
```diff
@@ -53,3 +54,123 @@ def angle_relative_to_horizontal(
...
+
+
+def _compute_nonwear_value_per_axis(
+    axis_acceleration_data: pl.Series, std_criteria: float, range_criteria: float
+) -> int:
+    """Helper function to calculate the nonwear criteria per axis.
+
+    Args:
+        axis_acceleration_data: The long window acceleration data for one axis.
+        It is polars.Series chunked into short windows where each row is a list of the
```
ReinderVosDeWael, line 162: Use additional indentation here to make clear where the arguments begin.

frey-perez (commented):
Largely looks good to me. Added some comments about doc strings, and minor changes regarding error handling to the main code to consider. Other than that I made some comments about how the testing should be handled that will probably merit further discussion. 

Overall it all looks healthy to me!

frey-perez (approved):

tests/unit/test_metrics.py
```diff
@@ -79,3 +97,154 @@ def test_angle_relative_to_horizontal(
...
+
+    test_resultx = metrics._compute_nonwear_value_per_axis(
+        acceleration_grouped["X"], std_criteria, range_criteria
+    )
+    test_resulty = metrics._compute_nonwear_value_per_axis(
+        acceleration_grouped["Y"], std_criteria, range_criteria
+    )
+    test_resultz = metrics._compute_nonwear_value_per_axis(
+        acceleration_grouped["Z"], std_criteria, range_criteria
+    )
```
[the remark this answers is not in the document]
Asanto32: Hmmm yeah you're right, I can just test one axis. I guess I wanted to make sure that each of the X, Y, and Z columns all worked. Note though that  `_compute_nonwear_value_per_axis` only returns 0 or 1, the  0, 1, 2, 3 values come from the sum of these, which is called in `_compute_nonwear_value_array`.

src/wristpy/processing/metrics.py
```diff
@@ -53,3 +54,184 @@ def angle_relative_to_horizontal(
...
+    """Set non_wear_flag based on accelerometer data.
+
+    This implements GGIR "2023" non-wear detection algorithm.
+    Briefly, the algorithm, creates a sliding window of long epoch length that steps
+    forward by the short epoch length.
+    It checks if the acceleration data in that long window, for each axis, meets certain
+    criteria thresholds to compute a non-wear value.
+    And then applies that non-wear value to all the short windows that make up the long
+    window. Additionally, for overlapping windows, the maximum of the non-wear value for
+    the overlaps is kept. Finally, there is a pass to find isolated "1s" in the non-wear
```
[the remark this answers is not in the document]
Asanto32: Haha will try to simplify, explaining what GGIR does is always a headache ha

src/wristpy/processing/metrics.py
```diff
@@ -53,3 +54,184 @@ def angle_relative_to_horizontal(
...
+
+
+    Returns:
+        A new Measurment instance with the non-wear flag and corresponding timestamps.
+    """
+    acceleration_grouped_by_short_window = _group_acceleration_data_by_time(
+        acceleration, short_epoch_length
+    )
+
+    n_short_epoch_in_long_epoch = int(long_epoch_length / short_epoch_length)
```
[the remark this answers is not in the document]
Asanto32: @ReinderVosDeWael thoughts on handling this. I can easily add a check/ Raise a ValueError if it isn't a multiple of short_epoch_length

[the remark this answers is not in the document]
ReinderVosDeWael: Is the result for short not exactly divisible by long valid? If it's valid, then why do we say it is ideally a multiple? If it's not valid, why do we allow it?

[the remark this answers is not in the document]
Asanto32: In GGIR, it is hardcoded to such that the short window is 900s and the long window is 3600s (the default values I set). Currently, it is valid and functional for any long_window that is a multiple of short_window. In theory, it could work with non-multiples? But there could be some issues at the end of the acceleration signal and I haven't tested the validity of this.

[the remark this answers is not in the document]
ReinderVosDeWael: Can we use an integer like `n_sub_windows` to compute the length of the 'short window' to side-step the user having to do math?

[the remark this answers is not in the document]
Asanto32: I am going to add this (and the equivalent test) for now. But maybe I will make a Task for future investigation

```   
 if long_epoch_length % short_epoch_length != 0:
        raise ValueError(
            "long_epoch_length must be an integer multiple of short_epoch_length"
        )
```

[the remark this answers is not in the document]
Asanto32: I don't know why I don't see these comments right away... but yeah that makes sense. What about the following modification:

We have the following `short_window_length: int` and `n_short_windows_in_long_epoch:int` instead of `long_epoch_length`. This way we don't need to worry about odd or prime `long_epoch_length`. We only use `long_epoch_length` to get the number of short windows that make up the long window.

tests/unit/test_metrics.py
```diff
@@ -79,3 +97,173 @@ def test_angle_relative_to_horizontal(
...
+
+
+@pytest.mark.parametrize(
+    "modifier, expected_result",
+    [
+        (1, 3),
+        (0, 0),
+    ],
+)
+def test_compute_nowear_value_array(
```
ReinderVosDeWael, line 190: Type nonwear :)

src/wristpy/processing/metrics.py
```diff
@@ -53,3 +54,184 @@ def angle_relative_to_horizontal(
...
+
+    for ones_index in nonwear_value_is_one:
+        if ones_index == 0 or ones_index == len(nonwear_value_array) - 1:
+            continue
+        if (nonwear_value_array[ones_index - 1] > 1) and (
+            nonwear_value_array[ones_index + 1] > 1
+        ):
+            nonwear_value_array[ones_index] = 2
+
+    return nonwear_value_array
```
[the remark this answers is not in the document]
Asanto32: Currently, the entire non-wear detection only takes a few seconds, even for the largest data set. But happy to try and optimize in a future version.

Asanto32:
@ReinderVosDeWael Updated to address all comments and discussion from yesterday. Not sure if you want to take a final look before merging. Note that the new `test_compute_nonwear_value_array` is a little intricate but it is to ensure we cover all possible outcomes (0, 1, 2, 3), and the best way I thought to do that was to manually modify one data axis at a time for the "special" cases to get the 1 and 2 non-wear values as outputs.

ReinderVosDeWael:
You should know that "test is a little intricate" is like a swan song to summon me :P.  Yeah, if you need such a large if statement to handle your parameterization, then you should not be parameterizing your test i.e. this one needs to be broken up. 

Will review in a sec.

src/wristpy/processing/metrics.py
```diff
@@ -53,3 +54,192 @@ def angle_relative_to_horizontal(
...
+        n_short_epoch_in_long_epoch: Number of short epochs that makeup one long epoch.
+        std_criteria: Threshold criteria for standard deviation.
+        range_criteria: Threshold criteria for range of acceleration.
+
+    Returns:
+        Non-wear value array.
+    """
+    nonwear_value_array = np.zeros(len(grouped_acceleration))
+
+    for window_n in range(len(grouped_acceleration) - n_short_epoch_in_long_epoch + 1):
```
ReinderVosDeWael, line 174: Can `len(...)` be puleld out of this statement and given a proper name? What are we looping over? `n_windows`?

src/wristpy/processing/metrics.py
```diff
@@ -53,3 +54,192 @@ def angle_relative_to_horizontal(
...
+
+    Returns:
+        Non-wear value for the axis.
+    """
+    axis_long_window_data = pl.concat(axis_acceleration_data, how="vertical")
+    axis_std = axis_long_window_data.std()
+    axis_range = axis_long_window_data.max() - axis_long_window_data.min()
+
+    criteria_boolean = (axis_std < std_criteria) & (axis_range < range_criteria)
+    return int(criteria_boolean)
```
ReinderVosDeWael, line 219: Why are we returning a boolean as an int?

src/wristpy/processing/metrics.py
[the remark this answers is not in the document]
ReinderVosDeWael: ~4-10 times faster depending on array size in my tests, plus more interpretable _to me_ (though that's subjective, of course):

```python
def _cleanup_isolated_ones_nonwear_value(nonwear_value_array: np.ndarray) -> np.ndarray:
    """Helper function to cleanup isolated ones in nonwear value array.
    This function finds isolated ones in the nonwear value array and
    sets them to 2 if they are surrounded by values > 1.
    Args:
        nonwear_value_array: The nonwear value array that needs to be cleaned up.
            It is a 1D numpy array.
    Returns:
        The modified nonwear value array.
    """
    left_neighbors = np.roll(nonwear_value_array, 1)
    right_neighbors = np.roll(nonwear_value_array, -1)

    condition = (left_neighbors > 1) & (right_neighbors > 1) & nonwear_value_array == 1
    condition[0] = False 
    condition[-1] = False 

    nonwear_value_array[condition] = 2

    return nonwear_value_array
```

tests/unit/test_metrics.py
```diff
@@ -79,3 +97,180 @@ def test_angle_relative_to_horizontal(
...
+@pytest.mark.parametrize(
+    "modifier, expected_result",
+    [
+        (1, 3),
+        (1, 2),
+        (1, 1),
+        (0, 0),
+    ],
+)
+def test_compute_nonwear_value_array(
```
ReinderVosDeWael, line 180: If each of your parameterizations requires a separate if-statement, its time to break up your test; I now have to understand the complete if-statement and its relationship to your parameter to understand what the test is doing.

src/wristpy/processing/metrics.py
ReinderVosDeWael: Python will actually do implicit conversions from bool to int when you try to do math with it e.g. `sum([True, True]) == 2`. So unless polars does some shenanigans here, you may be able to just remove the int cast without breaking anything else. 

If you do really need this as an int due to polars shenangians, I'd suggest converting it where you need it to make this clearer.

Asanto32:
Yeah, if you need such a large if statement to handle your parameterization, then you should not be parameterizing your test i.e. this one needs to be broken up.
> 
> Will review in a sec.

Makes sense. I realize now it can just be broken into individual case tests (well two cases are easily parameterized)

tests/unit/test_metrics.py
ReinderVosDeWael, line 180: Though in the larger scheme of things this parameterization may not be necessary at all. Changes in the modifier are already tested by your tests for `_compute_nonwear_value_per_axis` and this is the only location where these are used. 

Calling back to what I mentioned the other day about "scaling back tests to only what is needed", there are no branching paths in this function.  In other words, a single call can walk through every line of code in this function. Unless there's an implicit edge case (e.g. divide by zero shenanigans), wouldn't a single test suffice? 

Perhaps a good way to think about it is to ask "what things need to be tested, that are unique to `_compute_nonwear_value_array`".

ReinderVosDeWael (changes requested):

src/wristpy/processing/metrics.py
Asanto32: Yes, it is the total number of short windows. Will pull it out.

tests/unit/test_metrics.py
Asanto32: Yeah, initially I had this as one test. It walks through all the lines of code, but I wanted to ensure that I could get all potential results (0, 1, 2, 3). But maybe that isn't strictly a "unit" test? More of an integration/functional test?

ReinderVosDeWael: Its still a unit test but ask yourself this: in what way does the _behavior_ of the code change given a different numbered output? It's just taking a sum right i.e. which specific integer comes out isn't actually relevant.

src/wristpy/processing/metrics.py
Asanto32: Yeah, I just cast it because I do a polars.sum_horizontal() after and wasn't sure if it could handle booleans. Will check and convert if it works without casting.

src/wristpy/processing/metrics.py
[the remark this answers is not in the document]
Asanto32: Changed as suggested.

src/wristpy/processing/metrics.py
```diff
@@ -232,14 +233,15 @@ def _cleanup_isolated_ones_nonwear_value(nonwear_value_array: np.ndarray) -> np.
...
-    nonwear_value_is_one = np.where(nonwear_value_array == 1)[0]
-
-    for ones_index in nonwear_value_is_one:
-        if ones_index == 0 or ones_index == len(nonwear_value_array) - 1:
-            continue
-        if (nonwear_value_array[ones_index - 1] > 1) and (
-            nonwear_value_array[ones_index + 1] > 1
-        ):
-            nonwear_value_array[ones_index] = 2
+    nonwear_value_array = nonwear_value_array.astype(int)
```
ReinderVosDeWael, line 236: Something's going wrong here I think? What is the datatype of nonwear right now? If it's coming in as a boolean then this entire function is pointless (since its only 0's and 1's). If its already an int/float then why cast?

ReinderVosDeWael: Oh right if its a float then `== 1` can fail.

Asanto32: It is an np array of floats actually. It goes into polars.sum_horizontal() as an array of bool but after the code snippet below,  it is an array of floats. I was getting a type error ("`TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to...`") when doing the bitwise_and , prior to casting.

`        max_window_value = np.maximum(
            nonwear_value_array[window_n : window_n + n_short_epoch_in_long_epoch],
            np.repeat(calculated_nonwear_value, n_short_epoch_in_long_epoch),
        )
        
        nonwear_value_array[window_n : window_n + n_short_epoch_in_long_epoch] = (
            max_window_value
        )
`

ReinderVosDeWael (approved):

PR #57, opened 2024-07-08
Feat: Implement default sleep detection

### LINKED ISSUE #40: Task: Build SleepDetection algorithm.
### Description

The goal of this task is to build in the main GGIR sleep detection algorithm. The class should be built to allow for the implementation of more algorithms, but those are beyond the scope of this PR.

### Tasks

- [x] Create a SleepDetectionGgir class inside wristpy/processing/analytics.py
  - [x] It contains a `run()` method taking in (probably) `enmo`, `anglez`, `nonwear` and any other necessary data. 
  - **Note**: It may be given as many private functions (prefixed by `_`) as is needed for code organization. 

### Freeform Notes

Given the complexity of the GGIR algorithm (https://github.com/wadpac/GGIR/blob/master/R/g.part4.R), the exact implementation should be considered flexible.

Asanto32:
Resolves #40 
Adds the SleepDetection class to `analytics.py`, this function runs the default GGIR sleep detection algorithm and returns an instance of the `SleepWindow` data class that contains a list of the onset-wakeup time pairs for each SleepWindow.


src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,305 @@
...
+        calculated based on the overlap between the SPT windows and SIB periods.
+
+        Args:
+            method: the method to use for sleep detection. Currently only "GGIR" is
+                supported.
+
+        Returns:
+            A SleepWindow instance with sleep onset and wake up times.
+        """
+        if method == "GGIR":
```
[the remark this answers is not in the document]
ReinderVosDeWael: This may not be needed given earlier ABC comment.

ReinderVosDeWael (changes requested):

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,305 @@
...
+from typing import List, Optional, Tuple
+
+import numpy as np
+import polars as pl
+
+from wristpy.core import computations, models
+
+
+@dataclass
+class SleepWindow:
```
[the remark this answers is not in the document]
Asanto32: Yeah, I think that would be better, I think I misinterpreted the description in our wristpy architecture schmatic.

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,305 @@
...
+    Attributes:
+        onset: the predicted start time of the sleep window.
+        wakeup: the predicted end time of the sleep window.
+    """
+
+    onset: List[datetime]
+    wakeup: List[datetime]
+
+
+class SleepDetection:
```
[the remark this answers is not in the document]
Asanto32: Ok, yeah, there are two primary papers on it (one for SPTs and one for SIBs). I will include them. Would it make sense to include it under those specific methods or under the (new) general `GGIRSleepDetection` class

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,305 @@
...
+            threshold: the threshold for the distribution of z angle.
+
+        Returns:
+            A Measurement instance with values set to 1 indicating identified
+            SPT windows and corresponding time stamps.
+        """
+        anglez_abs_diff = self._compute_abs_diff_mean_anglez(anglez_data)
+        anglez_median_long_epoch = computations.moving_median(anglez_abs_diff, 300)
+        below_threshold = (
+            (anglez_median_long_epoch.measurements < threshold).astype(int).flatten()
```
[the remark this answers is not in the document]
Asanto32: Only used in the convolution. I guess this goes back to the same point with the nonwear detection and me not knowing that we could explicitly multiply boolean with int. Will remove.

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,305 @@
...
+
+    """
+
+    def __init__(
+        self,
+        anglez: models.Measurement,
+        nonwear: Optional[models.Measurement] = None,
+        method: str = "GGIR",
+    ) -> None:
+        """Initialize the SleepDetection class with default attributes."""
```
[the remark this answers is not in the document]
frey-perez: This came up in my review, whats the distinction in how we write these ? Should one be more verbose than the other ?

src/wristpy/processing/analytics.py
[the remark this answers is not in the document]
ReinderVosDeWael: Either is fine

src/wristpy/processing/analytics.py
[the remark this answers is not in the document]
Asanto32: anglez.measurements is 1D, but the return from `coputations.moving_median` is 2D.

I think it is probably this line:
`measurements=moving_median_df.drop("time").to_numpy()`

@frey-perez can maybe confirm?
Would this be something I modify even if it is 'out of scope' for the PR?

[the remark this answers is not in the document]
ReinderVosDeWael: Yeah that'd be out-of-scope here, though I do think we should look at moving_median altering the number of dimensions in a different issue.

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,305 @@
...
+        anglez_pl_df = pl.DataFrame(
+            {"time": anglez_abs_diff.time, "angz_diff": anglez_abs_diff.measurements}
+        )
+
+        anglez_pl_df = anglez_pl_df.with_columns(pl.col("time").set_sorted())
+        anglez_grouped_by_window_length = anglez_pl_df.group_by_dynamic(
+            index_column="time", every="5m"
+        ).agg([pl.all().exclude(["time"])])
+        flag = anglez_grouped_by_window_length["angz_diff"].map_elements(
+            lambda lst: all(x < threshold_degrees for x in lst)
```
frey-perez, line 141: is the loop necessary, wouldn't checking the max work ?

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,305 @@
...
+            If there is no overlap between spt_windows and sib_periods,
+            the onset and wakeup lists will be empty.
+        """
+        onset = []
+        wakeup = []
+        for spt in spt_periods:
+            min_onset = None
+            max_wakeup = None
+            for sib in sib_periods:
+                if sib[0] <= spt[1] and sib[1] >= spt[0]:
```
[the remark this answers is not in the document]
ReinderVosDeWael: nah i agree

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,305 @@
...
+        return sleep_idx_array
+
+    def _compute_abs_diff_mean_anglez(
+        self, anglez_data: models.Measurement, window_size_seconds: int = 5
+    ) -> models.Measurement:
+        """Helper function to compute the absolute difference of averaged anglez data.
+
+        Args:
+            anglez_data: the raw anglez data.
+                ##Note if we have access to the epoch1 anglez data this function will be
```
ReinderVosDeWael, line 253: Just noticed I missed this ##Note. If we need to keep track of this, put it in an issue along with the requirements for getting epoch1 anglez data.

tests/unit/test_analytics.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def test_find_periods(
+    sleep_detection: analytics.SleepDetection,
+) -> None:
+    """Test the _find_periods method."""
+    dummy_date = datetime(2024, 5, 2)
+    dummy_datetime_list = [dummy_date + timedelta(seconds=i) for i in range(10)]
+    test_time = pl.Series("time", dummy_datetime_list)
+    window_data = np.array([0, 1, 1, 1, 0, 1, 0, 0, 0, 1])
```
frey-perez, line 81: Worth testing the edge cases here ? All flagged/ all unflagged.

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,305 @@
...
+            spt_window_periods = self._find_periods(spt_window)
+            sib_window_periods = self._find_periods(sib_periods)
+            sleep_onset_wakeup = self._find_onset_wakeup_times(
+                spt_window_periods, sib_window_periods
+            )
+
+        return sleep_onset_wakeup
+
+    def _spt_window(
+        self, anglez_data: models.Measurement, threshold: float = 0.13
```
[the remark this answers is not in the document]
Asanto32: Yeah, I haven't tested how this value impacts the result. I'd like to leave it hardcoded for now, but in theory you are correct, someone could change it. Maybe I'll just make an issue to raise this as something to investigate in the future and then we can move it out as an arg in run like you said.

src/wristpy/processing/analytics.py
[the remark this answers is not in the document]
Asanto32: Ok will clarify the vars.

src/wristpy/processing/analytics.py
Asanto32: So `anglez_grouped_by_window_length["angz_diff"]` is a pl.Series where each element is a list. This loop checks whether all data points that makes up each list (I improved the naming convention of `x` and `lst` to hopefully clarify this) is below the threshold or not

FYI it looks like this now:
```
   flag = anglez_grouped_by_window_length["angz_diff"].map_elements(
            lambda grouped_list_data: all(
                list_data < threshold_degrees for list_data in grouped_list_data
            )
        )
 ```

File: tests/unit/test_analytics.py [hunk 11]
```diff
@@ -0,0 +1,163 @@
...
+
+def test_spt_window(sleep_detection: analytics.SleepDetection) -> None:
+    """Test the _spt_window method."""
+    sleep_detection.anglez.measurements = np.zeros(
+        len(sleep_detection.anglez.measurements)
+    )
+    half_long_block = 180
+    expected_length = int(len(sleep_detection.anglez.measurements) / 5) - 1
+    expected_result = np.zeros(expected_length)
+    expected_result[half_long_block:] = 1
```
frey-perez, line 102: Unnecessary alias of half_long_block?

src/wristpy/processing/analytics.py
ReinderVosDeWael: Unfortunately I'm not the Eye of Sauron; I don't see everything :'(

tests/unit/test_analytics.py
```diff
@@ -0,0 +1,163 @@
...
+    ]
+
+    result = sleep_detection._find_periods(window_measurement)
+
+    assert result == expected_result, f"Expected {expected_result}, but got {result}"
+
+
+def test_spt_window(sleep_detection: analytics.SleepDetection) -> None:
+    """Test the _spt_window method."""
+    sleep_detection.anglez.measurements = np.zeros(
```
frey-perez, line 96: I think complete activity and complete inactivity should be the two main edge cases to test.

tests/unit/test_analytics.py
```diff
@@ -0,0 +1,163 @@
...
+        wakeup=[dummy_date + timedelta(hours=4)],
+    )
+
+    result = sleep_detection._find_onset_wakeup_times(spt_periods, sib_periods)
+
+    assert result.onset == expected_output.onset
+    assert result.wakeup == expected_output.wakeup
+
+
+def test_run_sleep_detection(sleep_detection: analytics.SleepDetection) -> None:
```
frey-perez, line 155: Think i forgot to mention this in the last meeting, but I'll write this again, run doesn't actually do anything but call all the other functions in the right order, so really what kind of edge cases could there be?

Also eventually run will do one additional thing which is to pick the right method,  should we write the test for that now so we have it as this expands. @ReinderVosDeWael

tests/unit/test_analytics.py
```diff
@@ -0,0 +1,163 @@
...
+    sib_periods = [
+        (
+            dummy_date + timedelta(hours=2),
+            dummy_date + timedelta(hours=4),
+        )
+    ]
+    expected_output = analytics.SleepWindow(
+        onset=[dummy_date + timedelta(hours=2)],
+        wakeup=[dummy_date + timedelta(hours=4)],
+    )
```
frey-perez, line 147: How about adding the case of no overlap ?

frey-perez (commented):
Initial pass looks good, my comments are largely around extending some of the testing and some variable name / doc string stuff. Also sorry my comments are segmented was accidentally working in different windows!

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,305 @@
...
+
+        Args:
+            method: the method to use for sleep detection. Currently only "GGIR" is
+                supported.
+
+        Returns:
+            A SleepWindow instance with sleep onset and wake up times.
+        """
+        if method == "GGIR":
+            spt_window = self._spt_window(self.anglez)
```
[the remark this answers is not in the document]
Asanto32: We changed quite a bit of the class architecture with the new `AbstractSleepDetector` abstract class that should address these concerns. Will push that soon.

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,305 @@
...
+
+        return sleep_onset_wakeup
+
+    def _spt_window(
+        self, anglez_data: models.Measurement, threshold: float = 0.13
+    ) -> models.Measurement:
+        """Implement Heuristic distribution of z angle (HDCZA) for SPT window detection.
+
+        This function finds the absolute difference of the anglez data over 5s windows.
+        We find when the 5-minute rolling median of that difference is below a
```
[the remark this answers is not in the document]
Asanto32: Will do, I also have now included the references for HDCZA and SIB calculation.

tests/unit/test_analytics.py
Asanto32: This was also changed (new `find_periods`), and I don't think that needs to be tested anymore.

tests/unit/test_analytics.py
Asanto32: I thought it was more clear than `180` that comes out of nowhere?

tests/unit/test_analytics.py
Asanto32: Function has changed with new `GGIRSleepDetector`. I guess I'm just testing that we get the correct result when running the complete `sleep_detection.run_sleep_detection()`

tests/unit/test_analytics.py
Asanto32: Added this one using `modifier` in parametrized testing.

tests/unit/test_analytics.py
Asanto32: Technically this case is tested with `test_run_sleep_detection`. Should I explicitly add it in here @ReinderVosDeWael ?

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,330 @@
...
+        Args:
+            sleep_idx_array: the array of SPT windows.
+            gap_block: the length of the gap that defines sleep, default is 60 minutes.
+                The units are chunks of 5s.
+
+        Returns:
+            A numpy array with 1s indicating the identified SPT windows.
+        """
+        zeros_idx = np.where(sleep_idx_array == 0)[0]
+        n_blocks = np.split(zeros_idx, np.where(np.diff(zeros_idx) != 1)[0] + 1)
```
ReinderVosDeWael, line 266: This is a list of blocks, not a number of blocks, right?

Asanto32: Right, it's a list of arrays (break zeros_idx into a list of sub-arrays). Bad var name, will change

src/wristpy/processing/analytics.py
```diff
@@ -0,0 +1,330 @@
...
+                    and inactivity_bout[1] >= sleep_guide[0]
+                ):
+                    if min_onset is None or inactivity_bout[0] < min_onset:
+                        min_onset = inactivity_bout[0]
+                    if max_wakeup is None or inactivity_bout[1] > max_wakeup:
+                        max_wakeup = inactivity_bout[1]
+            if min_onset is not None and max_wakeup is not None:
+                sleep_windows.append(SleepWindow(onset=min_onset, wakeup=max_wakeup))
+
+        if not sleep_windows:
```
ReinderVosDeWael, line 215: This if-statement can be removed - you can return `[]`

tests/unit/test_analytics.py
```diff
@@ -0,0 +1,170 @@
...
+        (dummy_datetime_list[5], dummy_datetime_list[5]),
+        (dummy_datetime_list[9], dummy_datetime_list[9]),
+    ]
+
+    result = sleep_detection._find_periods(window_measurement)
+
+    assert result == expected_result, f"Expected {expected_result}, but got {result}"
+
+
+@pytest.mark.parametrize("modifier", [0, 1])
```
ReinderVosDeWael, line 95: Split this into two tests

ReinderVosDeWael (approved):

PR #66, opened 2024-07-19
Chore: Update to a simple readme

Asanto32:
Some initial changes for a new readme page. Copied style from other CMI repos. Logo taken from one of Freymon's slides.

Also, I'm sure the installation and Quick start  sections will change, especially after we implement a CLI/runner and more uses of the config.Settings. Just wanted something as a quick place holder.


ReinderVosDeWael (commented):
I'd have a good look at the logo - it has a partial alpha background which may make for some odd visuals depending on whether the browser/github support alpha. 

Also, if we're only uploading a .png as a "large" file then the following is overkill, but in case you don't know it yet, look up git-lfs. Its a tool for dealing with large files in git repositories.

nx10:
[No internet explorer 6 compatibility](https://caniuse.com/png-alpha)

ReinderVosDeWael:
Point being - the partial alpha (semi-circular on the left side) is gonna look weird on most browsers then

nx10:
If wristpy does not run under Windows XP what are we even doing. (Joking aside, I agree the logo needs its background removed)

Asanto32:
We can remove the logo for now (I was copying NiWrap and actfast aha). I just grabbed one of the example images that Freymon had used in a presentation. Probably Greg/John, etc. might have feedback and preferences on the logo anyway.

nx10:
It looks really nice if I remember next week I will remove the background and send it to you 👍

ReinderVosDeWael:
logo_alpha

ReinderVosDeWael (approved):

PR #73, opened 2024-07-24
Use numpy.inf for numpy 2.0

nx10:
Should fix tests failing in https://github.com/childmindresearch/wristpy/pull/69


ReinderVosDeWael (approved):
nice

PR #78, opened 2024-07-26
Feature/issue 44/test dependencies os workflow

Asanto32:
Attempting to add OS, python version, dependency version checks

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,18 +6,26 @@ on:
 
 jobs:
   unit:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: ['3.9', '3.10', '3.11', '3.12']
+        resolution: [lowest-direct, highest]
     steps:
     - uses: actions/checkout@v4
     - name: Install poetry
       run: pipx install poetry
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -6,18 +6,28 @@ on:
...
     steps:
     - uses: actions/checkout@v4
     - name: Install poetry
       run: pipx install poetry
     - uses: actions/setup-python@v5
       with:
-        python-version-file: pyproject.toml
-        cache: poetry
+        python-version: ${{ matrix.python_version }}
+        cache: pip
```
nx10, line 23:
```suggestion
        cache: poetry
```

nx10: Or does this not work with the version matrix?

Asanto32: Copied this from @ReinderVosDeWael implementation haha

Asanto32: I think because it uses `uv pip install` it is more efficient to cache pip?

ReinderVosDeWael: honestly :shrug: . This is a bit of an edge case for installing that allows us to do the `lowest-direct` and `highest` resolution.

ReinderVosDeWael (commented):
I'd suggest changing the on: push event back to only on a push to main. Adding 20-something duplicate test runners to every push that is also a PR is a bit... much.

ReinderVosDeWael (approved):

PR #80, opened 2024-07-30
Feature/issue 62/add logger

frey-perez:
Adding a logger to config.py.


frey-perez:
@ReinderVosDeWael Just noticed the low code coverage, it looks like setting

PR #83, opened 2024-07-30
Feature/issue 62/add logger

frey-perez:
Adding logger to config.py

### DIFF
--- a/src/wristpy/core/config.py
+++ b/src/wristpy/core/config.py
@@ -17,11 +17,11 @@ class Settings(pydantic_settings.BaseSettings):
 
 def get_logger() -> logging.Logger:
     """Gets the wristpy logger."""
-    if logging.getLogger("wristpy").hasHandlers():
-        return logging.getLogger("wristpy")
     logger = logging.getLogger("wristpy")
-    logger.setLevel(Settings().LOGGING_LEVEL)
+    if logger.handlers:
+        return logger
 
+    logger.setLevel(Settings().LOGGING_LEVEL)
     formatter = logging.Formatter(
         "%(asctime)s - %(name)s - %(levelname)s - %(filename)s:%(lineno)s - %(funcName)s - %(message)s",  # noqa: E501
     )
--- a/src/wristpy/core/config.py
+++ b/src/wristpy/core/config.py
@@ -1,5 +1,7 @@
 """Configuration module for wristpy."""
 
+import logging
+
 import pydantic_settings
 
 
@@ -9,3 +11,22 @@ class Settings(pydantic_settings.BaseSettings):
     LIGHT_THRESHOLD: float = 0.03
     MODERATE_THRESHOLD: float = 0.1
     VIGOROUS_THRESHOLD: float = 0.3
+
+    LOGGING_VERBOSITY: int = 20
+
+
+def get_logger() -> logging.Logger:
+    """Gets the wristpy logger."""
+    if logging.getLogger("wristpy").hasHandlers():
+        return logging.getLogger("wristpy")
+    logger = logging.getLogger("wristpy")
+    logger.setLevel(Settings().LOGGING_VERBOSITY)
+
+    formatter = logging.Formatter(
+        "%(asctime)s - %(name)s - %(levelname)s - %(filename)s:%(lineno)s - %(funcName)s - %(message)s",  # noqa: E501
+    )
+
+    handler = logging.StreamHandler()
+    handler.setFormatter(formatter)
+    logger.addHandler(handler)
+    return logger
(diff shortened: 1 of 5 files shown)


src/wristpy/core/config.py
```diff
@@ -9,3 +11,22 @@ class Settings(pydantic_settings.BaseSettings):
...
     VIGOROUS_THRESHOLD: float = 0.3
+
+    LOGGING_LEVEL: int = 20
+
+
+def get_logger() -> logging.Logger:
+    """Gets the wristpy logger."""
+    logger = logging.getLogger("wristpy")
+    if logger.handlers:
+        return logger
```
Asanto32, line 22: This line is never covered. I know when we spoke in the meeting you had to some intricate stuff  with the LogCaptureFixture to get by the if statement, but should we include a test for when the handler already exists?

src/wristpy/core/config.py
```diff
@@ -9,3 +11,22 @@ class Settings(pydantic_settings.BaseSettings):
     LIGHT_THRESHOLD: float = 0.03
     MODERATE_THRESHOLD: float = 0.1
     VIGOROUS_THRESHOLD: float = 0.3
+
+    LOGGING_LEVEL: int = 20
```
Asanto32, line 15: Given @ReinderVosDeWael comment on slack, should this be set to 10 for DEBUG?

tests/unit/test_logger.py
```diff
@@ -0,0 +1,19 @@
...
+def test_get_logger(caplog: pytest.LogCaptureFixture) -> None:
+    """Test the wristpy logger with level set to 20 (info)."""
+    logger = config.get_logger()
+
+    logger.debug("Debug message here.")
+    logger.info("Info message here.")
+    logger.warning("Warning message here.")
+
+    assert logger.getEffectiveLevel() == 20
+    assert "Debug message here" not in caplog.text
```
Asanto32, line 17: If we change to DEBUG level, I believe this will have to change to

Asanto32 (commented):
Some small comments here. My main question is whether we need to go populate the logger in the different modules (the primary candidates are most likely calibration, nonwear detection and sleep) in this PR? Or perhaps it will be mostly used in the runner/processor? We spoke about this in the meeting today but we weren't sure if that fits here or if it is a future task? Thoughts @ReinderVosDeWael

src/wristpy/core/config.py
frey-perez: In my conception of it, during the  normal course of running wristpy, the debug logs we write should be in the code, but not actually logging, so as not to flood the user with unnecessary information. When something goes wrong and we need to get in there and debug, you would switch the Settings().LOGGING_LEVEL to 10 yourself, and all those debug logs you wrote will now work. That was my reasoning for having 20 (INFO) as the default.

ReinderVosDeWael:
Regarding @Asanto32 's comment, yes I'd include adding the basic logging statements in this PR. I'd keep it to debug/error statements in some likely offenders for things going wrong in the future, and we can add info statements to the runners.

src/wristpy/processing/analytics.py
```diff
@@ -83,14 +85,17 @@ def run_sleep_detection(self) -> List[SleepWindow]:
...
+        logger.debug("Beginning sleep detection.")
         spt_window = self._spt_window(self.anglez)
         sib_periods = self._calculate_sib_periods(self.anglez)
         spt_window_periods = _find_periods(spt_window)
         sib_window_periods = _find_periods(sib_periods)
         sleep_onset_wakeup = self._find_onset_wakeup_times(
             spt_window_periods, sib_window_periods
         )
-
+        logger.debug(
```
ReinderVosDeWael, line 96: For all logging statements with variables in them, avoid f-strings (c.f. https://docs.astral.sh/ruff/rules/logging-f-string/ for an explanation of why). 

Personally, I prefer the *args method, but either *args or extra (both described in the above link) are fine. Optionally, you could enable Ruff's G-rules to be notified of this and several other issues.

src/wristpy/processing/calibration.py
```diff
@@ -201,6 +217,7 @@ def _chunked_calibration(
             CalibrationError: If all possible chunks have been used and the calibration
                 process fails to get below the `min_calibration_error` threshold.
         """
+        logger.debug("running chunked calibration.")
```
ReinderVosDeWael, line 220: Minor thing; but you're mixing capitalized and non-capitalized logging statements. I'd suggest making a consistent choice here.

tests/unit/test_logger.py
ReinderVosDeWael: That's fine - if we do the test will fail and it can be modified at that point in time.

ReinderVosDeWael (changes requested):
Looks good overall -- couple of minor comments. I'd also suggest adding logging to the file reader.

ReinderVosDeWael:
What's the current status on this? Needs another review, still in the works?

frey-perez:
> What's the current status on this? Needs another review, still in the works?
Ah forgot to ping, yes it's ready for review.

PR #85, opened 2024-08-01
Fix: Add field validator for unique time in Measurement

### LINKED ISSUE #33: Task: Modify Measurement class in models.py to ensure that all time points are unique.
### Description

Currently, we only verify that that the time attribute is sorted, a pl.Datetime, and not empty. However, if all the timestamps are identical that is considered sorted. This is both a problem for downstream computations and also indicative of some corrupt data collection.

### Tasks

- [ ] Add field validator to ensure all entries of Measurement.time are unique
- [ ] Verify any unit tests impacted by this change.

### Freeform Notes

_No response_

Asanto32:
Resolves #33 
- added validator for unique time for Measurement object 
- added unit test support

### DIFF
--- a/src/wristpy/core/models.py
+++ b/src/wristpy/core/models.py
@@ -40,7 +40,8 @@ class Measurement(BaseModel):
     def validate_time(cls, v: pl.Series) -> pl.Series:
         """Validate the time series.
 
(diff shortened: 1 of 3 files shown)


frey-perez (approved):
Looks good

PR #86, opened 2024-08-02
Feat: Adding method to remove sleep windows that overlap with non-wear

### LINKED ISSUE #65: Task: Add a method to remove SleepWindows that overlap with non-wear time
### Description

There should not be any non-wear/sleepWindow overlap (assuming the non-wear detection is trust worthy). 

#TODO decide how and where this should be implemented (user chosen/config setting?)

### Tasks

- [ ] Create method that, if config setting `remove_non_wear_from_sleep == True`, removes sleep windows that overlap with non-wear time

### Freeform Notes

Sample code on how I am doing it currently.

```python 
nonwear_periods = sleep_detector_class._find_periods(non_wear_array)
filtered_sleep_windows = []
for sleep_window in sleep_windows:
    if not any((nonwear_period[0] <= sleep_window.onset <= nonwear_period[1] or 
                nonwear_period[0] <= sleep_window.wakeup <= nonwear_period[1]) or
                (sleep_window.onset <= nonwear_period[0] and sleep_window.wakeup >= nonwear_period[1])
               for nonwear_period in nonwear_periods):
        filtered_sleep_windows.append(sleep_window)

### PR DESCRIPTION
**Pull request by Asanto32**
Resolves #65 and #60 
Adding in a function that removes the sleep windows that have any overlap with the non-wear times. 
Moving the _find_periods out of the GGIRSleepDetection class so it can be used across SleepDetectors and by this function

### DIFF
--- a/src/wristpy/processing/analytics.py
+++ b/src/wristpy/processing/analytics.py
@@ -85,8 +85,8 @@ class GGIRSleepDetection(AbstractSleepDetector):
         """
         spt_window = self._spt_window(self.anglez)
         sib_periods = self._calculate_sib_periods(self.anglez)
-        spt_window_periods = self._find_periods(spt_window)
-        sib_window_periods = self._find_periods(sib_periods)
+        spt_window_periods = _find_periods(spt_window)
+        sib_window_periods = _find_periods(sib_periods)
         sleep_onset_wakeup = self._find_onset_wakeup_times(
             spt_window_periods, sib_window_periods
         )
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Asanto32** [comment]
Implementation of removing the sleep windows that overlap with non-wear periods

File: src/wristpy/processing/analytics.py [hunk 1]
```diff
@@ -308,41 +308,81 @@ def _compute_abs_diff_mean_anglez(
...
+
+
+def remove_nonwear_from_sleep(
+    non_wear_array: models.Measurement,
+    sleep_windows: List,
+) -> List[SleepWindow]:
+    """Remove non-wear periods from sleep windows.
+
+    This function finds all the non-wear periods and removes any sleep windows that have
+    any overlap with the non-wear periods (including non-wera completely
```
frey-perez, line 356: typo here "non-wera"

src/wristpy/processing/analytics.py
```diff
@@ -308,41 +308,81 @@ def _compute_abs_diff_mean_anglez(
...
+    within the sleep window).
+
+    Args:
+        non_wear_array: The non-wear array generated from metrics.detect_nonwear
+        sleep_windows (list): The list of sleep windows, where the entries are
+            instances of the SleepWindow class.
+        anglez (Measurement): The anglez data.
```
frey-perez, line 363: as per the style guide do we need the typing in the docs here?

tests/unit/test_analytics.py
```diff
@@ -88,7 +86,74 @@ def test_find_periods(
...
+        - where the non-wear period is within the sleep window.
+    """
+    dummy_date = datetime(2024, 5, 2)
+    sleep_windows = []
+    sleep_windows.append(
+        analytics.SleepWindow(
+            onset=dummy_date + timedelta(hours=2),
+            wakeup=dummy_date + timedelta(hours=4),
+        )
+    )
```
frey-perez, line 119: this is likely superfluous but would it be worth adding just one (non overlapping) window just to "prove" it's not just throwing out everything when there's an overlap.

frey-perez (approved):
Looks good to me, a few minor comments, nothing major.

tests/unit/test_analytics.py
```diff
@@ -88,7 +86,74 @@ def test_find_periods(
...
+
+    assert result == expected_result, f"Expected {expected_result}, but got {result}"
+
+
+@pytest.mark.parametrize(
+    "non_wear_array, expected_result",
+    [
+        (np.array([0, 1, 1, 1, 0]), []),
+        (np.array([0, 0, 0, 1, 1]), []),
+        (np.array([0, 0, 1, 0, 0]), []),
```
frey-perez, line 99: We discussed it in private but just adding this comment as a reminder to maybe change the wording of the docs below to better reflect these tests, and also change the third array so that the non wear is at n = 3.

src/wristpy/processing/analytics.py
Asanto32: Yeah, good catch. Got lazy with auto-complete.

tests/unit/test_analytics.py
Asanto32: Yeah, do you mean like in the `test_remove_nonwear_periods_no_overlap`? It is the test right after this one

src/wristpy/processing/analytics.py
```diff
@@ -308,41 +308,81 @@ def _compute_abs_diff_mean_anglez(
...
+    all_periods = single_periods + block_periods
+    all_periods.sort()
+
+    return all_periods
+
+
+def remove_nonwear_from_sleep(
+    non_wear_array: models.Measurement,
+    sleep_windows: List,
+) -> List[SleepWindow]:
```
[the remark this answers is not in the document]
frey-perez: my mistake, I think what i was thinking about was [PEP 585](https://peps.python.org/pep-0585/), but turns out the ' | ' operator isnt supported until 3.10 so Union is still needed.

tests/unit/test_analytics.py
frey-perez: In my head I was thinking more of demonstrating the function was able to _selectively_ remove the bad sleep period while still retaining the good one, but thats just smooshing the two tests together now that I type it out, so go ahead and ignore this ha.

src/wristpy/processing/analytics.py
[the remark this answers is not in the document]
ReinderVosDeWael: Yup, or if you're lazy like me add `from __future__ import annotations` to the top of the file. Either works, unless you're using Pydantic in which case the future workaround does **not** work.

src/wristpy/processing/analytics.py
```diff
@@ -308,41 +308,80 @@ def _compute_abs_diff_mean_anglez(
...
-    def _find_periods(
-        self, window_measurement: models.Measurement
-    ) -> List[Tuple[datetime, datetime]]:
-        """Find periods where window_measurement is equal to 1.
 
-        This is a helper function for the _find_onset_wakeup_times to
-        find periods where either the spt_window or sib_periods are equal to 1.
+def _find_periods(
+    window_measurement: models.Measurement,
+) -> List[Tuple[datetime, datetime]]:
```
ReinderVosDeWael, line 314: datetime is imported from where? :)

ReinderVosDeWael (changes requested):

src/wristpy/processing/analytics.py
```diff
@@ -308,41 +308,80 @@ def _compute_abs_diff_mean_anglez(
...
+    ]
+    all_periods = single_periods + block_periods
+    all_periods.sort()
+
+    return all_periods
+
+
+def remove_nonwear_from_sleep(
+    non_wear_array: models.Measurement,
+    sleep_windows: List,
```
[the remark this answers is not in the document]
Asanto32: Huh I forgot it there but it was in the description in the docstring. Fixed

tests/unit/test_analytics.py
```diff
@@ -88,7 +86,74 @@ def test_find_periods(
...
+
+
+def test_remove_nonwear_periods_no_overlap() -> None:
+    """Test the _remove_nonwear_from_sleep method.
+
+    This test is for the case where the non-wear period does not overlap
+    with the sleep window.
+    """
+    dummy_date = datetime(2024, 5, 2)
+    sleep_windows = []
```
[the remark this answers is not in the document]
Asanto32: Yes, fixed for both.

src/wristpy/processing/analytics.py
Asanto32: I switched it everywhere so now at the import level it is just
`import datetime`

and for the class definition it is `datetime.datetime`. Is this correct? the datetime.datetime confuses me aha

src/wristpy/processing/analytics.py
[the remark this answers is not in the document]
Asanto32: Oh, now I remember. It's because SleepWindow can be `datetime | List` so the current version leads to a mypy error later. Added a check to ensure that it is datetime first (aka not empty sleepwindow)

tests/unit/test_analytics.py
```diff
@@ -88,7 +86,72 @@ def test_find_periods(
...
+    non_wear_array: np.ndarray, expected_result: List
+) -> None:
+    """Test the _remove_nonwear_from_sleep method.
+
+    This test is for the following cases:
+        - where the non-wear period overlaps with sleep onset.
+        - where the non-wear period overlaps with sleep wakeup.
+        - where the non-wear period is within the sleep window.
+    """
+    dummy_date = datetime(2024, 5, 2)
```
ReinderVosDeWael, line 112: where did this datetime come from? :)

ReinderVosDeWael (approved):

tests/unit/test_analytics.py
Asanto32: fixed (we might need to do a thorough review through all unit tests at some point)

PR #91, opened 2024-08-23
Bugfix: Change _find_long_blocks to catch leading edge

### LINKED ISSUE #88: Bug: analytics._fill_short_blocks does not seem to be functioning properly in testing
### Description

The `_fill_short_blocks` function is used to fill in short gaps in the SPT window. While the test is passing, it doesn't seem to function correctly with real data (gaps that should be filled are not, leading to multiple SPT windows that have gaps < 1hr, the default gap length)

### Steps to Reproduce

For example: Analyzing this file (1897__104846_2024-08-05 11-29-13.bin) and then looking at the SPT_windows in the generation of sleep windows, there are gaps < 1 hour (using all default settings and running `processor.process_file()` in the adhd_summer_camp_anlalysis module.

### Expected Results

There should not be any gaps < 1hour between the end of a SPT window and the start of the next.

### Actual Results

There are short gaps that aren't filled in. 

### Additional Notes

Trying to modify the unit test to see where the issue may be.

### LINKED ISSUE #90: Bug: Sleep onset is delayed by half_block_length
### Description

The `_find_long_blocks` function in `analytics.py` uses a convolution kernel to find periods of sleep that are > `long_block_length`, however, since this is a convolution the returned index of that overlap is the central point. We will implement a single pointer method, similar to `_fill_short_blocks` to find the first instance of that overlap.

### Steps to Reproduce

This is evident even in the unit test, where the expected_result should not be shorter than incoming array `below_threshold`

### Expected Results

We will modify the `_find_long_blocks` so that the returned array has sleep (SPT_window) start at the beginning of the overlap.

### Actual Results

Currently the start of the `SPT_window` is `half_block_length` later than it should be.

### Additional Notes

_No response_

Asanto32:
Resolves #90 
Resolves #88 

Changed the `_find_long_blocks` to use single pointer similar to `_fill_short_blocks`, this eliminates the issue that arose with using a centered convolution (center point of convolution kernel instead of leading edge of spt_window detected).

Note the modification to the `test_spt_window` compared to the previous test. We initialize the anglez.measurements to be a random number, this can actually cause the test to fail if the block immediately prior to our 'simulated sleep' looks sufficiently like sleep to create a connected longer window. The edge cases are now tested directly in `test_find_long_blocks`.

----------
Modified _find_long_blocks to use single pointer implementation to find chunks of 1s. Modified tests so that test_find_long_blocks now finds start of 1s in middle and at ending of below_threshold. Modified test_spt_window to simpler implementation to remove odd cases where random num generator of anglez caused unpredictable sleep detection at leading edge

### DIFF
--- a/src/wristpy/processing/analytics.py
+++ b/src/wristpy/processing/analytics.py
@@ -102,10 +102,9 @@ class GGIRSleepDetection(AbstractSleepDetector):
         We find the 5-minute rolling median of that difference.
         Next, we find when that 5-minute median is above a specified threshold, taken as
         the new default value from the GGIR implementation of the HDCZA algorithm. This
-        represent non-sleep candidates. The logical not of this is the sleep candidate.
+        represents non-sleep candidates. The logical not of this is the sleep candidate.
         We then find long blocks (30 minutes) when the threshold criteria is met.
-        Any gaps in SPT windows that are less than a specified window length
-        (default 60 minutes) are filled.
+        Any gaps in SPT windows that are less than 60 minutes are filled.
 
         Args:
             anglez_data: the raw anglez data, calculated from calibrated acceleration.
@@ -121,17 +120,25 @@ class GGIRSleepDetection(AbstractSleepDetector):
               using an accelerometer without sleep diary. Sci Rep 8, 12975 (2018).
               https://doi.org/10.1038/s41598-018-31266-z
         """
+        long_epoch_median = 300
+        long_block = 360
+        short_block_gap = 720
+
         anglez_abs_diff = self._compute_abs_diff_mean_anglez(anglez_data)
-        anglez_median_long_epoch = computations.moving_median(anglez_abs_diff, 300)
+        anglez_median_long_epoch = computations.moving_median(
+            anglez_abs_diff, long_epoch_median
+        )
         non_sleep_candidates = (
             anglez_median_long_epoch.measurements >= threshold
         ).flatten()
 
         sleep_candidates = np.logical_not(
-            self._fill_false_short_blocks(non_sleep_candidates, 360)
+            self._fill_false_blocks(non_sleep_candidates, long_block)
         )
 
-        sleep_idx_array_filled = self._fill_false_short_blocks(sleep_candidates, 720)
+        sleep_idx_array_filled = self._fill_false_blocks(
+            sleep_candidates, short_block_gap
+        )
 
         return models.Measurement(
             measurements=sleep_idx_array_filled, time=anglez_median_long_epoch.time
@@ -225,7 +232,7 @@ class GGIRSleepDetection(AbstractSleepDetector):
 
         return sleep_windows
 
-    def _fill_false_short_blocks(
+    def _fill_false_blocks(
         self, boolean_array: np.ndarray, gap_block: int
     ) -> np.ndarray:
         """Helper function to fill gaps in SPT window that are less than gap_blocks.
@@ -237,9 +244,8 @@ class GGIRSleepDetection(AbstractSleepDetector):
         if that value is less than the gap_block, we fill in with ones.
 
         Args:
-            boolean_array: A generic boolean array that the array of SPT windows.
-            gap_block: the length of the gap that needs to be filled, for SPT windows,
-                the defaul block length is assumed to be
+            boolean_array: A generic boolean array, typically the SPT window.
+            gap_block: the length of the gap that needs to be filled.
 
         Returns:
             A numpy array with 1s, typically for identified SPT windows.
@@ -258,7 +264,7 @@ class GGIRSleepDetection(AbstractSleepDetector):
                 continue
 
             if n_zeros < gap_block:
-                boolean_array[sleep_array_idx - n_zeros : sleep_array_idx] = 1
+                boolean_array[sleep_array_idx - n_zeros : sleep_array_idx] = True
             n_zeros = 0
 
         return boolean_array
--- a/src/wristpy/processing/analytics.py
+++ b/src/wristpy/processing/analytics.py
@@ -253,11 +253,9 @@ class GGIRSleepDetection(AbstractSleepDetector):
             if spt_value:
                 n_ones += 1
                 continue
-            if (not spt_value) & (n_ones >= block_length):
+            elif n_ones >= block_length:
                 sleep_idx_array[spt_array_idx - n_ones : spt_array_idx] = 1
-                n_ones = 0
-            if not spt_value:
-                n_ones = 0
+            n_ones = 0
 
         if n_ones >= block_length:
             sleep_idx_array[-n_ones:] = 1
--- a/src/wristpy/processing/analytics.py
+++ b/src/wristpy/processing/analytics.py
@@ -225,25 +225,42 @@ class GGIRSleepDetection(AbstractSleepDetector):
     ) -> np.ndarray:
         """Helper function to find long blocks where SPT window is true.
 
-        This function uses the convolution of a kernel of 1s, of length block_length,
-        to find the continuous long blocks where SPT window is true. Where the
-        convolution value is == long_block_length that implies an overlap between the
-        kernel and the threshold signal of length long_block.
+        This function finds the first non-zero in below_threshold array, if there are
+        none, we return the initial array.
+        We then iterate over the array and count all the ones between zeros
+        (skipping the first 1), if that value is >= long_block we fill in with ones.
 
         Args:
             below_threshold: the 5-minute rolling median of the anglez difference
-                that is true when below the cutoff threshold.
+                that is true when below the cutoff threshold, 5s temporal resolution.
             block_length: the length of the long block that defines sleep, default is
                 30 minutes. (360 chunks of 5s)
 
         Returns:
             A numpy array with 1s indicating the identified SPT windows.
         """
-        kernel = np.ones(block_length, dtype=int)
-        convolved = np.convolve(below_threshold, kernel, mode="same")
-        long_blocks_idx = np.where(convolved == block_length)[0]
+        n_ones = 0
         sleep_idx_array = np.zeros(len(below_threshold))
-        sleep_idx_array[long_blocks_idx] = 1
+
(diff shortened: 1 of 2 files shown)


src/wristpy/processing/analytics.py
```diff
@@ -225,25 +225,42 @@ def _find_long_blocks(
...
+        )
+        if first_one_idx is None:
+            return below_threshold
+
+        for spt_array_idx in range(first_one_idx, len(below_threshold)):
+            spt_value = below_threshold[spt_array_idx]
+            if spt_value:
+                n_ones += 1
+                continue
+            if (not spt_value) & (n_ones >= block_length):
```
ReinderVosDeWael, line 256: `spt_value` is zero by definition due to the above if-statement.

src/wristpy/processing/analytics.py
```diff
@@ -225,25 +225,42 @@ def _find_long_blocks(
...
+                sleep_idx_array[spt_array_idx - n_ones : spt_array_idx] = 1
+                n_ones = 0
+            if not spt_value:
```
ReinderVosDeWael, line 259: spt_value is zero by definition; this if statement can be removed entirely.

ReinderVosDeWael:
So what's the intended difference between fill_short_blocka and fill_long_blocks?

src/wristpy/processing/analytics.py
```diff
@@ -225,25 +225,42 @@ def _find_long_blocks(
...
             A numpy array with 1s indicating the identified SPT windows.
         """
-        kernel = np.ones(block_length, dtype=int)
-        convolved = np.convolve(below_threshold, kernel, mode="same")
-        long_blocks_idx = np.where(convolved == block_length)[0]
+        n_ones = 0
         sleep_idx_array = np.zeros(len(below_threshold))
-        sleep_idx_array[long_blocks_idx] = 1
+
+        first_one_idx = next(
```
ReinderVosDeWael, line 245: Is it intentional that leading zeros are never filled but trailing zeros are?

Asanto32:
> So what's the intended difference between fill_short_blocka and find_long_blocks? Could these not essentially be merged with some minor overhead?

Hmmm, maybe. I was just following the ordered steps in [the GGIR algorithm](https://www.nature.com/articles/s41598-018-31266-z). First it finds "long blocks" of potential sleep, these are continuous blocks > 30 mins (default value). Then it fills in any gaps between these potential sleep windows that are 60 mins (default value) or less. This creates the final `spt_window`, which is the guider for where we look for sustained inactivity bouts (SIB). The overlap of the two makes the sleep window.

src/wristpy/processing/analytics.py
Asanto32: Changed logic for this section.

src/wristpy/processing/analytics.py
ReinderVosDeWael: Right, so both functions are effectively taking a list of booleans and you invert the boolean if their sequence is less than a certain length right? find_long_block turns 1 into 0s and fill_short_block 0 into 1?

ReinderVosDeWael: Got that comment's location mixed up with the main chain

ReinderVosDeWael:
So looking at this I'm still not sure I understand why this can't be the same function.

src/wristpy/processing/analytics.py
```diff
@@ -225,25 +225,44 @@ def _find_long_blocks(
     ) -> np.ndarray:
         """Helper function to find long blocks where SPT window is true.
 
-        This function uses the convolution of a kernel of 1s, of length block_length,
-        to find the continuous long blocks where SPT window is true. Where the
-        convolution value is == long_block_length that implies an overlap between the
-        kernel and the threshold signal of length long_block.
+        This function finds the first non-zero in below_threshold array, if there are
```
ReinderVosDeWael, line 228: This is definitely going into documentation overkill territory -- let the code speak for itself more.

Asanto32:
This now Resolves #88 (there was an indentation error causing short blocks after a large gap to not be filled)

src/wristpy/processing/analytics.py
```diff
@@ -220,70 +225,43 @@ def _find_onset_wakeup_times(
...
-        convolved = np.convolve(below_threshold, kernel, mode="same")
-        long_blocks_idx = np.where(convolved == block_length)[0]
-        sleep_idx_array = np.zeros(len(below_threshold))
-        sleep_idx_array[long_blocks_idx] = 1
-
-        return sleep_idx_array
-
-    def _fill_short_blocks(
-        self, sleep_idx_array: np.ndarray, gap_block: int = 720
+    def _fill_false_short_blocks(
```
ReinderVosDeWael, line 228: Since the distinction between short/long block doesnt exist anymore (at least not in code) - I'd just rename it to _fill_false_blocks. Especially since we're now also using this for "long" blocks

src/wristpy/processing/analytics.py
```diff
@@ -220,70 +225,43 @@ def _find_onset_wakeup_times(
...
         We then iterate over the array and count every zero between ones
-        (skipping the first 1),
+        (skipping the leading zeros),
         if that value is less than the gap_block, we fill in with ones.
 
         Args:
-            sleep_idx_array: the array of SPT windows.
-            gap_block: the length of the gap that defines sleep, default is 60 minutes.
-                The units are chunks of 5s.
+            boolean_array: A generic boolean array that the array of SPT windows.
```
ReinderVosDeWael, line 240: Rephrase; I'm not sure what you mean here.

src/wristpy/processing/analytics.py
```diff
@@ -220,70 +225,43 @@ def _find_onset_wakeup_times(
...
+            gap_block: the length of the gap that needs to be filled, for SPT windows,
```
ReinderVosDeWael, line 241: "the defaul block length is assumed to be"...? There is no default right?

src/wristpy/processing/analytics.py
```diff
@@ -220,70 +225,43 @@ def _find_onset_wakeup_times(
...
+        for sleep_array_idx in range(first_one_idx, len(boolean_array)):
+            sleep_value = boolean_array[sleep_array_idx]
             if not sleep_value:
                 n_zeros += 1
                 continue
 
             if n_zeros < gap_block:
-                sleep_idx_array[sleep_array_idx - n_zeros : sleep_array_idx] = 1
-                n_zeros = 0
+                boolean_array[sleep_array_idx - n_zeros : sleep_array_idx] = 1
```
ReinderVosDeWael, line 261: Since we're expecting only boolean arrays, I'd use True instead of 1 here; I'm not entirely sure how numpy is going to cast this otherwise.

tests/unit/test_analytics.py
```diff
@@ -20,35 +20,20 @@ def sleep_detection() -> analytics.GGIRSleepDetection:
...
-
-def test_fill_short_blocks(
-    sleep_detection: analytics.GGIRSleepDetection,
-) -> None:
-    """Test the _fill_short_blocks method."""
+    """Test the _fill_false_short_blocks method."""
     gap_block = 3
-    sleep_idx_array = np.array([0, 0, 0, 1, 1, 0, 0, 1])
-    expected_result = np.array([0, 0, 0, 1, 1, 1, 1, 1])
+    sleep_idx_array = np.array([1, 0, 1, 0, 0, 0, 1, 1, 0, 1, 0, 1])
```
ReinderVosDeWael, line 33: We expect arrays of True/False, so use those here too.

src/wristpy/processing/analytics.py
```diff
@@ -100,8 +100,9 @@ def _spt_window(
 
         This function finds the absolute difference of the anglez data over 5s windows.
         We find the 5-minute rolling median of that difference.
-        Next, we find what that 5-minute median is below a specified threshold, taken as
-        new default value from the GGIR implementation of the HDCZA algorithm.
+        Next, we find when that 5-minute median is above a specified threshold, taken as
+        the new default value from the GGIR implementation of the HDCZA algorithm. This
+        represent non-sleep candidates. The logical not of this is the sleep candidate.
```
ReinderVosDeWael, line 105: small typo: "This representS"

src/wristpy/processing/analytics.py
```diff
@@ -122,11 +123,15 @@ def _spt_window(
         """
         anglez_abs_diff = self._compute_abs_diff_mean_anglez(anglez_data)
         anglez_median_long_epoch = computations.moving_median(anglez_abs_diff, 300)
```
ReinderVosDeWael, line 125: In general, if you use numbers as input for a function without a variable name, I'd recommend using a keyword argument e.g. `moving_median(anglez_abs_diff, epoch_length=300)`. As an aside, the same recommendation holds for True/False.

ReinderVosDeWael: The same goes for the numbers in fill_short_blocks. This should also make it a bit more evident that these are different units.

src/wristpy/processing/analytics.py
```diff
@@ -100,8 +100,9 @@ def _spt_window(
...
         This function finds the absolute difference of the anglez data over 5s windows.
         We find the 5-minute rolling median of that difference.
-        Next, we find what that 5-minute median is below a specified threshold, taken as
-        new default value from the GGIR implementation of the HDCZA algorithm.
+        Next, we find when that 5-minute median is above a specified threshold, taken as
+        the new default value from the GGIR implementation of the HDCZA algorithm. This
+        represent non-sleep candidates. The logical not of this is the sleep candidate.
         We then find long blocks (30 minutes) when the threshold criteria is met.
         Any gaps in SPT windows that are less than a specified window length
         (default 60 minutes) are filled.
```
ReinderVosDeWael, line 108: It's 60 minutes, period. It's not a 'default' as there is no option to change it.

ReinderVosDeWael (approved):
A bunch of minor comments; overall just have a good look at the docstrings again. I'll approve it already since none of these comments are on any of the functionality.

Issue #94, opened 2024-08-27
Task: write runner 


frey-perez:
### Description

Write a runner that will call the argparser from the cli module, and use those arguments to call the run function from the orchestrator module. This runner will allow wristpy to execute the main data processing pipeline from the command line. 

### Tasks

- [ ] write runner module. 
   - will call cli module
   - will feed args to orchestrator module.

frey-perez:
This runner() module will be combined with the cli module for simplicity.

Issue #71, opened 2024-07-24
Task: Add a Python entry point


ReinderVosDeWael:
### Description

Similar to #41 - Add an entrypoint for Python. This can largely use the same code bar input argument parsing.

### Tasks

- [ ] Add a Python-based runner. 
    - [ ] For arguments, this should take the exact same as the CLI.

Issue #41, opened 2024-05-22
Task: Write the CLI argument parser


ReinderVosDeWael:
### Description

Many end-users will use wristpy via a shell. The CLI should allow for this. 

### Tasks

- [ ] Create parse_cli_args() function in wristpy/core/cli.py
  - [ ] This will use argparse to evaluate the following arguments:
    - [ ] the input file 
    - [ ] the output file
    - [ ] start (int|None) for start date (optional, default None)
    - [ ] end (int|None) for end date (optional, default None)
    - [ ] Optional arguments for core/config.py settings
    - [ ] It should return an argparse.NameSpace

Running the code based on these arguments is left to the orchestrator and out of scope for this PR.

PR #95, opened 2024-08-30
Feature/issue 71/add python entry point

frey-perez:
Addressing issue #71, creates an orchestrator module that will allow the main wristpy pipeline to work via python script. The run function will return a Results object with all the relevant data, and will save data into two(raw_time, and epcoh1) csv or parquet files when given an output via the run function or the save_reuslts method. 

This orchestrator is also intended to work together with the arparser in issue #41 , within the runner addressed in issue #94, to allow wristpy to run in the command line.


src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+) -> np.ndarray:
+    """Formats sleep windows into an array for saving.
+
+    Args:
+        sleep_windows: The list of time stamp pairs indicating periods of sleep.
+        reference_measure: The measure from which the temporal resolution will be taken.
+            Any epoch1 measure can be used, but enmo is used for consistency.
+
+    Returns:
+        1-D binary np.ndarray, with 1 indicating sleep. Will be of the same length as
```
ReinderVosDeWael, line 129: Nope - its the same length as the reference measure.

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+def format_sleep_data(
+    sleep_windows: List[analytics.SleepWindow], reference_measure: models.Measurement
+) -> np.ndarray:
+    """Formats sleep windows into an array for saving.
+
+    Args:
+        sleep_windows: The list of time stamp pairs indicating periods of sleep.
+        reference_measure: The measure from which the temporal resolution will be taken.
```
ReinderVosDeWael, line 125: This function doesn't use epoch1, that's decided by the caller.

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+            "nonwear": nonwear_data.measurements.astype(np.int64),
+            "time": nonwear_data.time,
+        }
+    ).set_sorted("time")
+
+    nonwear_upsample = nonwear_df.upsample(time_column="time", every="5s").fill_null(
+        strategy="forward"
+    )
+
+    end_sequence = pl.repeat(
```
ReinderVosDeWael, line 169: Why are we just extending to the length of the reference? What if the reference doesn't have the same delta as we hardcode in the upsample?

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+
+    return sleep_array
+
+
+def format_nonwear_data(
+    nonwear_data: models.Measurement, reference_measure: models.Measurement
+) -> np.ndarray:
+    """Format nonwear data into an array for saving.
+
+    Args:
```
ReinderVosDeWael, line 148: See docstring comments above.

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+    full_nonwear_array = pl.concat([nonwear_upsample["nonwear"], end_sequence])
+
+    return full_nonwear_array.to_numpy()
+
+
+def run(
+    input: pathlib.Path,
+    output: Optional[pathlib.Path] = None,
+    settings: config.Settings = config.Settings(),
+    calibrator: calibration.Calibration = calibration.Calibration(),
```
ReinderVosDeWael, line 184: I'd use a literal string here (`Literal["ggir"]`, or whatever an appropriate name is). Part of the goal of an orchestrator is to make things as easy on the user as possible; this set-up forces them to know where to import calibrators from in the first place.

One asterisk: if your intent is to allow users to provide their own calibrators, then we should discuss. In that particular case a setup very similar to this would be OK.

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+    detect_nonwear_kwargs: Optional[Dict[str, Any]] = None,
```
ReinderVosDeWael, line 185: See typing.Unpack https://docs.python.org/3/library/typing.html#typing.Unpack for how to type this.

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+        calibrator: The calibrator to be used on the input data.
+        detect_nonwear_kwargs: The arguments to the nonwear function delivered as a
+            dictionary. Arguements are short_epoch_length, n_short_epoch_in_long_epoch,
+            std_criteria, range_criteria.
+
+    Returns:
+        All calculated data in a save ready format as a Results object.
+
+    """
+    watch_data = readers.read_watch_data(input)
```
ReinderVosDeWael, line 205: When dealing with user input, fail early. We can already check at the start whether the output extension is in a list of accepted extensions.

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+    )
+    full_nonwear_array = pl.concat([nonwear_upsample["nonwear"], end_sequence])
+
+    return full_nonwear_array.to_numpy()
+
+
+def run(
+    input: pathlib.Path,
+    output: Optional[pathlib.Path] = None,
+    settings: config.Settings = config.Settings(),
```
ReinderVosDeWael, line 183: (read the comment on calibrator first).

This is the one spot I'd allow an exception for having an import from elsewhere in the package, if only for the simple reason that replicating its interface is an absolute pain :P

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+            std_criteria, range_criteria.
+
+    Returns:
+        All calculated data in a save ready format as a Results object.
+
+    """
+    watch_data = readers.read_watch_data(input)
+    try:
+        calibrated_acceleration = calibrator.run(watch_data.acceleration)
+    except Exception as e:
```
ReinderVosDeWael, line 208: This is dangerous; too many things can go wrong in a generic exception to just continue without thinking about what happened. Any reason why we don't provide the option to _not_ use a calibration and just fail if calibration fails? If this happens so often that we need an explicit edge case, then its a condemnation of our calibration function that needs fixing.

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+        print("Calibration FAILED:", str(e))
+        print("Proceeding without calibration.")
+        calibrated_acceleration = watch_data.acceleration
+    enmo = metrics.euclidean_norm_minus_one(calibrated_acceleration)
+    enmo_epoch1 = computations.moving_mean(enmo)
```
ReinderVosDeWael, line 213: So I know there's a default on moving_mean, but now that I read the usage here I'd be inclined to remove the default entirely and enforce an explicit window everywhere so it's always clear what the epoch is. @Asanto32 thoughts?

src/wristpy/core/orchestrator.py
ReinderVosDeWael: Oh wait, this isnt a real kwargs (no double asterisk), thats some confusing variable naming (: . I'd just replace entirely with a TypedDict and rename it to something like nonwear_settings. 

Alternatively, should these settings be ported to the overall settings @Asanto32 ?

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+        pairs, which are used to create a binary array in epoch1 time. Two files will
+        be saved, one of each temporal resolution. This means if output is entered as
+        /path/to/file/file.csv,The file names will be labeled as file_epoch1.csv and
+        file_raw_time.csv in the path/to/file directory.
+
+        Args:
+            output: The path and file name of the data to be saved. as either a csv or
+                parquet files.
+
+        Returns:
```
ReinderVosDeWael, line 53: No need for a returns block if nothing is returned

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+        sleep_windows=sleep_windows,
+        physical_activity_levels=physical_activity_levels,
+        sleep_windows_epoch1=sleep_array,
+        nonwear_epoch1=nonwear_epoch1,
+    )
+    if output is not None:
+        try:
+            results.save_results(output=output)
+        except (InvalidFileTypeError, FileNotFoundError) as e:
+            print(e)
```
ReinderVosDeWael, line 257: Rebase this branch on the logging branch so you can use the logger for this.

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+            InvalidFileTypeError: If the output file path ends with any extension other
+                than csv or parquet.
+        """
+        if output.suffix not in [".csv", ".parquet"]:
+            raise InvalidFileTypeError(
+                f"The extension: {output.suffix} is not supported.",
+                "Please save the file as .csv or .parquet",
+            )
+
+        results_epoch1_time = self._results_to_dataframe()
```
ReinderVosDeWael, line 66: Could it be rewritten as a single function that just takes the field names of the fields that should be saved( e.g. `_result_to_dataframe(["enmo", "anglez", ...])`)

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+
+        if data_source is None:
+            raise ValueError(f"{data_source} is None, can't construct time column.")
+
+        time_data = data_source.time
+
+        results_dataframe = pl.DataFrame({"time": time_data})
+
+        for field_name, field_value in self:
+            if field_value is not None:
```
ReinderVosDeWael, line 105: Why would a field be none?

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+from wristpy.processing import analytics, calibration, metrics
+
+
+class InvalidFileTypeError(Exception):
+    """Wristpy cannot save in the given file type."""
+
+    pass
+
+
+class Results(BaseModel):
```
ReinderVosDeWael, line 21: Keep an eye on being consistent with import styling i.e. this would be pydantic.BaseModel if you use the module imports everywhere else.

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+    Args:
+        sleep_windows: The list of time stamp pairs indicating periods of sleep.
+        reference_measure: The measure from which the temporal resolution will be taken.
+            Any epoch1 measure can be used, but enmo is used for consistency.
+
+    Returns:
+        1-D binary np.ndarray, with 1 indicating sleep. Will be of the same length as
+            the timestamps in the epoch1_measure.
+    """
+    sleep_array = np.zeros(len(reference_measure.time))
```
ReinderVosDeWael, line 132: docstrings claims this is a binary, but this is a float64

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+            Any epoch1 measure can be used, but enmo is used for consistency.
+
+    Returns:
+        1-D binary np.ndarray, with 1 indicating nonwear. Will be of the same length as
+            the timestamps in the reference_measure.
+
+    """
+    nonwear_df = pl.DataFrame(
+        {
+            "nonwear": nonwear_data.measurements.astype(np.int64),
```
ReinderVosDeWael, line 160: Docstring claims you're outputting a binary, but you're building an int64. Does polars allow casting as boolean?

src/wristpy/core/orchestrator.py
ReinderVosDeWael: Either way, removing the default would be outside this PR.

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,259 @@
...
+        time=enmo_epoch1.time,
+    )
+    nonwear_epoch1 = models.Measurement(
+        measurements=format_nonwear_data(
+            nonwear_data=non_wear_array, reference_measure=enmo_epoch1
+        ),
+        time=enmo_epoch1.time,
+    )
+
+    results = Results(
```
ReinderVosDeWael, line 242: Since creating epoch1 measures is always just a moving mean, could we not move this into the save_results step? i.e. somehting like `result.save_results(output, epoch=5)` or `result.save_results(output, epoch=None)`.

This would also make saving other epochs require no code duplication.

tests/unit/test_orchestrator.py
```diff
@@ -0,0 +1,208 @@
+"""Test the orchestrator.py module."""
```
ReinderVosDeWael, line 1: Since there's quite a lot of comments on the code itself, I'll delay reviewing tests until later.

src/wristpy/core/orchestrator.py
ReinderVosDeWael: See first the comment on epochs at the end of this file. That comment may make this one unnecessary.

ReinderVosDeWael (changes requested):

src/wristpy/core/orchestrator.py
Asanto32: Yeah, those all have default values in `detect_nonwear` and I would agree that they could be modified in the `config.Settings`. Especially as it looks like the `std_criteria` and `range_criteria` will need to be set by the readers (based on GENEActiv vs Actigraph). I don't expect the `short_epoch_length` and `n_short_epoch_in_long_epoch` to be touched.

src/wristpy/core/orchestrator.py
Asanto32: One thing that is changing (potentially) is that there will now be `calibration.Calibration.method = 'GGIR'` to use the GGIR method. Otherwise all other options will go to the new default (zscore for initial guess + scipy.optimize)

src/wristpy/core/orchestrator.py
frey-perez: I was getting mypy errors here as I have the fields as Optional, and defaulting to None, so explicitly ruling that out here was how I addressed it.

src/wristpy/core/orchestrator.py
frey-perez: This was me being silly, when i said binary here i meant in terms of the possible values only being 1s and 0s, not the data type. I'll remove that to prevent confusion.

src/wristpy/core/orchestrator.py
frey-perez: I like making these settings, I'll make that change, and make it so the detect_nonwear function gets its default values from Settings.

src/wristpy/core/orchestrator.py
Asanto32: My concern is that it seems more and more that the 'typical end user' wants something as simple as `run wristpy /path/to/directory/actigraphy_files/` and then they wouldn't have to look at anything. A big benefit in comparison to the current gold standard is that it would actually succeed in batch processing an entire directory without a ~25% error rate (trying to run GGIR on the HBN dataset was a nightmare, many unclear errors that just cause the program to crash). Especially given the discussion with Connor yesterday, I would lean towards continuing with no calibration if our calibration fails, and including the reason for failure in some `wristpy_runtime_metadata` report?
Alternatively, I guess we can develop our own `wristpy_run_directory` that can handle this nicely?

src/wristpy/core/orchestrator.py
Asanto32: Sorry I missed this. But we can remove the default, I think I was setting it up that way since the 5s "epoch_1" timing was so commonly used in a lot of the GGIR functionality

poetry.lock
Asanto32, line 1: These changes and the ones in `pyproject.toml`, `calibration` and `test_calibration` will presumably be covered by the calibration PR merge (hopefully no conflicts)

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,266 @@
...
+        if output.suffix == ".csv":
+            results_dataframe.write_csv(output, separator=",")
+            logger.debug("results saved in: %s", output)
+        elif output.suffix == ".parquet":
+            results_dataframe.write_parquet(output)
+            logger.debug("results saved in: %s", output)
+
+
+def validate_output(output: pathlib.Path) -> None:
+    """Validates that the output path exists and is a valid format.
```
Asanto32, line 66: Do we want to error or just make a new directory/path if it doesn't exist yet?

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+def run(
+    input: pathlib.Path,
+    output: Optional[pathlib.Path] = None,
+    settings: config.Settings = config.Settings(),
+    calibrator: Union[
+        None,
+        Literal["ggir", "gradient"],
+    ] = "gradient",
```
Asanto32, line 159: Did we say the default would be `None` in the end?

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,266 @@
...
+    epoch_length: Union[int, None] = 5,
+) -> Results:
+    """Runs wristpy.
```
Asanto32, line 162: Maybe we can be a bit more descriptive? "Runs the main processing steps of wristpy to generate data for analysis of physical activity and sleep data" or something along those lines?

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,266 @@
...
+    anglez = metrics.angle_relative_to_horizontal(calibrated_acceleration)
+    if epoch_length is not None:
+        enmo = computations.moving_mean(enmo, epoch_length=epoch_length)
+        anglez = computations.moving_mean(anglez, epoch_length=epoch_length)
+    if input.suffix == ".bin":
+        non_wear_array = metrics.detect_nonwear(
+            calibrated_acceleration, range_criteria=0.5
+        )
+    else:
+        non_wear_array = metrics.detect_nonwear(calibrated_acceleration)
```
Asanto32, line 226: To discuss (or outside this PR?) how to handle `.gt3x` vs `.bin`

tests/unit/local.ipynb
Asanto32, line 1: This file should be removed?

tests/smoke/test_orchestrator_smoke.py
```diff
@@ -0,0 +1,76 @@
...
+    assert (tmp_path / "good_file.csv").exists()
+    assert isinstance(results, orchestrator.Results)
+    assert results.enmo is not None
+    assert results.anglez is not None
+    assert results.nonwear_epoch is not None
+    assert results.sleep_windows_epoch is not None
+    assert results.physical_activity_levels is not None
+
+
+def test_bad_calibrator(test_file: pathlib.Path) -> None:
```
Asanto32, line 51: Add tests for the good calibration options?

tests/smoke/test_orchestrator_smoke.py
Asanto32, line 1: Out of curiosity is there a reason we have two separate files to test the orchestrator?

tests/unit/local.ipynb
frey-perez: Oops! My mistake.

tests/smoke/test_orchestrator_smoke.py
frey-perez: The unit tests cover the small helper functions as usual but the run function needed a smoke test, this was my first stab at one. Still some overlap with the unit tests, so not quite sure i nailed it.

tests/smoke/test_orchestrator_smoke.py
frey-perez: Technically it's defaulting to using one of the good calibrators right now so this is being done in the "happy path" test. Should I add a test where this is more explicit @ReinderVosDeWael ?

src/wristpy/core/orchestrator.py
frey-perez: good point, added a bit more description.

tests/smoke/test_orchestrator_smoke.py
ReinderVosDeWael: (Responding before reading the smoke test) : The goal of a smoke test is to detect major issues early on, not to really test every unit of code. With larger runner functions like the orchestrator, a 'proper' unit test is difficult to write, hence why I recommended @frey-perez write a test that more generically tests that the code runs, without error, and without obvious failures. 

Typically smoke and unit tests would also run at different parts of the pipeline, but at the scales that we work at that's overkill.

src/wristpy/core/orchestrator.py
```diff
@@ -0,0 +1,271 @@
...
+
+        """
+        logger.debug("Saving results.")
+        validate_output(output=output)
+
+        results_dataframe = pl.DataFrame({"time": self.enmo.time})
+
+        for field_name, field_value in self:
+            results_dataframe = results_dataframe.with_columns(
+                pl.Series(field_name, field_value.measurements)
```
ReinderVosDeWael, line 54: What do you think of this? Pick whichever is simpler to you. 

```python
results_dataframe = pl.DataFrame(
    {"time": self.enmo.time} |
    {name: value.measurements for name, value in self}
)

ReinderVosDeWael (changes requested):

ReinderVosDeWael (approved):

PR #98, opened 2024-09-03
Feat/issue 8/add gradient descent to calibration

Asanto32:
Adding a new default calibration method to solve the closest_point_fit optimization. The function will use the scale and offset to z-score the no_motion_data as the initial guess to `scipy.optimize` to minimize the error to the unit sphere. 
`GGIR` must now be called explicitly in `Calibration.method`

### DIFF
--- a/src/wristpy/processing/calibration.py
+++ b/src/wristpy/processing/calibration.py
@@ -68,7 +68,7 @@ class Calibration:
                 calibrate on entire dataset.
             method: Define the method used to find the calibration scaling and offset
                 that minimizes the calibration error. Current options are:
-                    'Gradient', 'GGIR'
+                    'Zscore', 'GGIR', default is 'Gradient'.
             min_acceleration: Minimum acceleration for sphere criteria.
             min_calibration_hours: Minimum hours of data required for calibration.
             min_standard_deviation: Minimum standard deviation for no-motion detection.
@@ -285,6 +285,10 @@ class Calibration:
             linear_transformation = self._closest_point_fit(
                 no_motion_data=no_motion_data
             )
+        elif self.method == "ZScore":
+            linear_transformation = self._closest_point_fit_zscore(
+                no_motion_data=no_motion_data
+            )
         else:
             linear_transformation = self._closest_point_fit_gradient_descent(
                 no_motion_data=no_motion_data
@@ -433,6 +437,26 @@ class Calibration:
 
         return LinearTransformation(scale=scale, offset=offset)
 
+    def _closest_point_fit_zscore(
+        self, no_motion_data: np.ndarray
+    ) -> LinearTransformation:
+        """Find linear transformation parameters that minimzes distance to unit sphere.
+
+        This function finds the linear transformation scale and offset parameters based
+        on the normalized z-score of the no motion data.
+
+        Args:
+            no_motion_data: The acceleration data during periods of no
+                motion, in order to determine scale and offset.
+
+        Returns:
+            A LinearTransformation object with scale and offset parameters to be applied
+            to acceleration data for calibration.
+        """
+        scale = 1 / (np.sqrt(3) * np.std(no_motion_data, axis=0))
+        offset = -np.mean(no_motion_data, axis=0) * scale
+        return LinearTransformation(scale=scale, offset=offset)
+
     def _closest_point_fit_gradient_descent(
         self, no_motion_data: np.ndarray
     ) -> LinearTransformation:
@@ -445,8 +469,7 @@ class Calibration:
         under 1g acceleartion due to Earth's gravity).
 
         Args:
-            no_motion_data: The acceleration data during periods of no
-                motion, in order to determine scale and offset.
+            no_motion_data: The acceleration data during periods of no motion.
 
         Returns:
             A LinearTransformation object with scale and offset parameters to be applied
--- a/src/wristpy/processing/calibration.py
+++ b/src/wristpy/processing/calibration.py
@@ -38,7 +38,7 @@ class SphereCriteriaError(LoggedException):
 
 
 class CalibrationError(LoggedException):
-    """Was not able to lower calibration below error threshold."""
+    """Calibration failed to converge or is not below the error threshold."""
 
     pass
 
@@ -562,7 +562,7 @@ class ConstrainedMinimizationCalibration(AbstractCalibrator):
             optimal_scale = result.x[:3]
             optimal_offset = result.x[3:]
         else:
-            raise CalibrationError("Optimization failed")
+            raise CalibrationError("Optimization failed.")
 
         no_motion_calibrated = (no_motion_data * result.x[:3]) + result.x[3:]
         cal_error_end = np.mean(abs(np.linalg.norm(no_motion_calibrated, axis=1) - 1))
@@ -570,9 +570,11 @@ class ConstrainedMinimizationCalibration(AbstractCalibrator):
         if cal_error_end >= self.max_calibration_error:
             cal_error_initial = np.mean(abs(np.linalg.norm(no_motion_data, axis=1) - 1))
             logger.debug(
-                "Calibration error could not be sufficiently minimized. "
-                f"Initial Error: {cal_error_initial}, Final Error: {cal_error_end}, "
-                f"Error threshold: {self.max_calibration_error}"
+                "Calibration error could not be sufficiently minimized."
+                "Initial Error: %s,  Final Error: %s, Error threshold: %s.",
+                cal_error_initial,
+                cal_error_end,
+                self.max_calibration_error,
             )
             raise CalibrationError(
                 "Calibration error could not be sufficiently minimized."
@@ -592,7 +594,7 @@ def _extract_no_motion(
     no_motion_threshold and a mean value < 2. These epochs are determined to be
     the periods where the accelerometer was influenced by no motion beyond the
     force of gravity. If periods of no motion are found, the ndarray is returned,
-    to be fit to the idealized unit sphere for the purposes of calibration
+    to be fit to the idealized unit sphere for the purposes of calibration.
 
     Args:
         acceleration: the accelerometer data containing x,y,z axis
--- a/src/wristpy/processing/calibration.py
+++ b/src/wristpy/processing/calibration.py
@@ -89,7 +89,7 @@ class AbstractCalibrator(abc.ABC):
 
 
 class GgirCalibration(AbstractCalibrator):
-    """Implements the GGIR calibration on accelerometer data..
+    """Implements the GGIR calibration on accelerometer data.
 
     This class implements methods for autocalibrating accelerometer data using either
     entire dataset or subsets, as determined by the settings. Depending on the settings
@@ -100,12 +100,12 @@ class GgirCalibration(AbstractCalibrator):
     Attributes:
         chunked: If true will preform calibration on subsets of data if false will
             calibrate on entire dataset.
-        min_acceleration: Minimum acceleration for sphere criteria.
+        min_acceleration: Minimum acceleration for sphere criteria, in g-force.
         min_calibration_hours: Minimum hours of data required for calibration.
-        min_standard_deviation: Minimum standard deviation for no-motion detection.
+        no_motion_threshold: Minimum standard deviation for no-motion detection.
         max_iterations: Maximum number of iterations for optimization.
         error_tolerance: Tolerance for optimization convergence.
-        min_calibration_error: Threshold for calibration error.
+        max_calibration_error: Threshold for the maximum allowable calibration error.
 
     """
 
@@ -114,10 +114,10 @@ class GgirCalibration(AbstractCalibrator):
         chunked: bool = False,
         min_acceleration: float = 0.3,
         min_calibration_hours: int = 72,
-        min_standard_deviation: float = 0.013,
+        no_motion_threshold: float = 0.013,
         max_iterations: int = 1000,
         error_tolerance: float = 1e-10,
-        min_calibration_error: float = 0.01,
+        max_calibration_error: float = 0.01,
     ) -> None:
         """Initializes class.
 
@@ -132,10 +132,10 @@ class GgirCalibration(AbstractCalibrator):
             min_calibration_hours: The minimum amount of data in hours needed to
                 preform calibration. Default is 72. If chunked calibration is selected
                 this will be the size of the initial subset taken for calibration. If
-                error has not been reduced below the min_calibration_error value, an
+                error has not been reduced below the max_calibration_error value, an
                 additional 12 hours will be taken until all data is used or calibration
                 is successful.
-            min_standard_deviation: The standard deviation critieria used to select
+            no_motion_threshold: The standard deviation critieria used to select
                 portions of the data with no movement. Rolling windows with standard
                 deviations below this value will be determined as "still". This value is
                 likely to be different between devices. Default is 0.013g. This value
@@ -147,8 +147,8 @@ class GgirCalibration(AbstractCalibrator):
             error_tolerance: Tolerated level of error, when the
                 closest_point_fit method arrives at this value or better, the process
                 ends. Default is 1e-10. Generally should be left at this value.
-            min_calibration_error: Minimum acceptable error. If calibration can
-                not reach this threshold it will error.
+            max_calibration_error: Maximum acceptable calibration error. If calibration
+                cannot reach this threshold it will error.
 
         Returns:
             None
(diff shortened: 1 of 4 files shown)


src/wristpy/processing/calibration.py
```diff
@@ -417,6 +433,59 @@ def _closest_point_fit(self, no_motion_data: np.ndarray) -> LinearTransformation
...
+        the no_motion_data.
+
+        Args:
+            no_motion_data: The acceleration data during periods of no motion.
+
+        Returns:
+            A LinearTransformation object with scale and offset parameters to be applied
+            to acceleration data for calibration.
+
+        Raises:
```
ReinderVosDeWael, line 456: Don't we already have a custom CalibrationError for this?

src/wristpy/processing/calibration.py
```diff
@@ -417,6 +433,59 @@ def _closest_point_fit(self, no_motion_data: np.ndarray) -> LinearTransformation
...
+        """Find linear transformation parameters that minimzes distance to unit sphere.
+
+        This function implements the scipy optimize.minimze function to find the
+        optimial scale and offset parameters that will minimze the error function,
+        which is defined as the distance of the no_motion acceleration data
+        from the unit sphere (where the unit sphere represents the ideal data points
+        under 1g acceleartion due to Earth's gravity). The initial guess for the
+        scale and offset are chosen by the equivalent transformation that would z-score
+        the no_motion_data.
+
```
ReinderVosDeWael, line 448: I'd explicitly document the reason for the bounds, as this may look odd to an uninformed reader.

src/wristpy/processing/calibration.py
```diff
@@ -272,7 +281,14 @@ def _calibrate(self, acceleration: models.Measurement) -> LinearTransformation:
             2014 Oct 1;117(7):738-44. doi: 10.1152/japplphysiol.00421.2014.
         """
         no_motion_data = self._extract_no_motion(acceleration=acceleration)
-        linear_transformation = self._closest_point_fit(no_motion_data=no_motion_data)
+        if self.method == "GGIR":
```
ReinderVosDeWael, line 284: We should discuss together with @frey-perez whether this is the best method for implementing this, rather than a separate class per calibrator. The reason being that it'd play more nicely with the proposal in #95 to allow custom calibrators into the orchestrator.

tests/unit/test_calibration.py
```diff
@@ -123,14 +123,49 @@ def test_closest_point_fit() -> None:
...
+        linear_transform.offset, expected_offset, atol=1e-3
+    ), f"Offset is {linear_transform.offset} expected {expected_offset})"
+
+
+def test_closest_point_fit_gradient_value_error() -> None:
+    """Test closest point fit gradient raises ValueError."""
+    data = np.random.randn(1000, 3)
+    calibrator = calibration.Calibration(max_iterations=0)
+
+    with pytest.raises(ValueError):
```
ReinderVosDeWael, line 154: Is this a ValueError from us or from scipy? c.f. https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions for how to confirm its actually ours, assuming that is the intent.

tests/unit/test_calibration.py
```diff
@@ -212,7 +247,7 @@ def test_run_calibration() -> None:
     """Testing run function when chunked = False."""
     scale = np.array([1.1, 1.01, 0.9])
     offset = np.array([0.1, 0.2, 0.1])
-    dummy_no_motion = np.random.randn(1000, 3) - 0.5
+    dummy_no_motion = np.random.randn(1000, 3) - 0.2
```
ReinderVosDeWael, line 250: Why is this value changed?

tests/unit/test_calibration.py
```diff
@@ -143,7 +178,7 @@ def test_calibration_successful() -> None:
     offset = np.array([0.1, 0.2, 0.1])
     expected_scale = 1 / scale
     expected_offset = -offset / scale
-    dummy_no_motion = np.random.randn(1000, 3) - 0.5
+    dummy_no_motion = np.random.randn(1000, 3)
```
ReinderVosDeWael, line 181: Why is this value changed?

tests/unit/test_calibration.py
```diff
@@ -123,14 +123,49 @@ def test_closest_point_fit() -> None:
     ), f"Offset is {linear_transform.offset} expected {expected_offset})"
 
 
+def test_closest_point_fit_gradient() -> None:
```
ReinderVosDeWael, line 126: note that the ggir functions should be renamed to indicate that they're ggir, as opposed to gradient desc.

ReinderVosDeWael (changes requested):

src/wristpy/processing/calibration.py
```diff
@@ -65,6 +66,9 @@ class Calibration:
     Attributes:
             chunked: If true will preform calibration on subsets of data if false will
                 calibrate on entire dataset.
+            method: Define the method used to find the calibration scaling and offset
+                that minimizes the calibration error. Current options are:
+                    'GGIR', or the default 'Gradient'.
```
clane9, line 71: Small nit re the terminology, `scipy.minimize` doesn't technically use gradient descent by default, but rather LBFGS (a better algorithm in this context). Might consider calling the method `'minimize'` or `'scipy'` or sth.

src/wristpy/processing/calibration.py
```diff
@@ -417,6 +433,59 @@ def _closest_point_fit(self, no_motion_data: np.ndarray) -> LinearTransformation
...
+        data_range = no_motion_data.max() - no_motion_data.min()
+        scale_bounds = [(0.1, None)] * 3
+        offset_bounds = [(-0.5 * data_range, 0.5 * data_range)] * 3
+        bounds = scale_bounds + offset_bounds
+
+        result = optimize.minimize(
+            get_distance_to_unit_sphere,
+            initial_guess,
+            options={"maxiter": self.max_iterations},
+            bounds=bounds,
```
clane9, line 478: If you decide to add the manual gradient computation, need to also set `jac=True`.

```suggestion
            bounds=bounds,
            jac=True,
```

src/wristpy/processing/calibration.py
```diff
@@ -417,6 +433,59 @@ def _closest_point_fit(self, no_motion_data: np.ndarray) -> LinearTransformation
...
+            ValueError if the optimization has failed.
+        """
+        start_scale = 1 / (np.sqrt(3) * np.std(no_motion_data, axis=0))
+        start_offset = -np.mean(no_motion_data, axis=0) * start_scale
+
+        def get_distance_to_unit_sphere(params: list[float]) -> float:
+            scale = params[:3]
+            offset = params[3:]
+            distances = np.linalg.norm((no_motion_data * scale) + offset, axis=1) - 1
+            return np.sum(distances**2)
```
clane9, line 466: I would recommend computing the gradient yourself rather than relying on scipy numerical gradient estimation. Will be more precise and reduce function evals by ~10x.

```suggestion
            transformed = scale * no_motion_data + offset
            norms = np.linalg.norm(transformed, axis=-1)
            deltas = norms - 1.0
            obj = (deltas ** 2).mean()

            scaled_transformed = (deltas[:, None] / norms[:, None]) * transformed
            grad_offset = scaled_transformed.mean(axis=0)
            grad_scale = (scaled_transformed * no_motion_data).mean(axis=0)
            grad = np.concatenate([grad_scale, grad_offset])
            return obj, grad
```

src/wristpy/processing/calibration.py
```diff
@@ -417,6 +433,59 @@ def _closest_point_fit(self, no_motion_data: np.ndarray) -> LinearTransformation
...
+
+        initial_guess = np.concatenate([start_scale, start_offset])
+        data_range = no_motion_data.max() - no_motion_data.min()
+        scale_bounds = [(0.1, None)] * 3
+        offset_bounds = [(-0.5 * data_range, 0.5 * data_range)] * 3
```
clane9, line 471: These offset bounds won't work in the (rare) case of a large ground truth offset but small data range. A better option might be to center the offset bounds around the initial offset guess.

```suggestion
        offset_bounds = [(val - 0.5 * data_range, val + 0.5 * data_range) for val in start_offset]
```

src/wristpy/processing/calibration.py
ReinderVosDeWael: huh, I thought LBFGS was a gradient descent algorithm, TIL

clane9: Tbf, it is definitely a nit comment 😅. It is true that LBFGS uses only first order information. But the step direction it computes is better than the naive negative gradient direction (using an approximate Hessian).

tests/unit/test_calibration.py
Asanto32: Accidentally saved a version when I was playing around with values for the tests when initially adding z-score method + gradient descent. Switched it back now. (Same for above)

tests/unit/test_calibration.py
Asanto32: Relic from when you wrote the function, will adjust to match our custom CalibrationError

src/wristpy/processing/calibration.py
```diff
@@ -462,3 +439,192 @@ def _get_sampling_rate(timestamps: pl.Series) -> int:
...
+            acceleration=acceleration,
+            min_standard_deviation=self.min_standard_deviation,
+        )
+        linear_transformation = self._closest_point_fit(no_motion_data=no_motion_data)
+
+        no_motion_calibrated = (
+            no_motion_data * linear_transformation.scale
+        ) + linear_transformation.offset
+
+        cal_error_initial = np.round(
```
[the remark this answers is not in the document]
ReinderVosDeWael: Maybe this is also available in the results object and you could pull this logging down to the fit function?

src/wristpy/processing/calibration.py
```diff
@@ -462,3 +439,192 @@ def _get_sampling_rate(timestamps: pl.Series) -> int:
...
+
+        Args:
+            no_motion_data: The acceleration data during periods of no motion.
+
+        Returns:
+            A LinearTransformation object with scale and offset parameters to be applied
+            to acceleration data for calibration.
+
+        Raises:
+            Calibration Error: If the optimization process fails to converge.
```
ReinderVosDeWael, line 553: No space in CalibrationError

tests/unit/test_calibration.py
```diff
@@ -123,27 +123,80 @@ def test_closest_point_fit() -> None:
...
+
+    assert np.allclose(
+        linear_transform.scale, expected_scale, atol=1e-3
+    ), f"Scale is {linear_transform.scale} expected {expected_scale}"
+    assert np.allclose(
+        linear_transform.offset, expected_offset, atol=1e-3
+    ), f"Offset is {linear_transform.offset} expected {expected_offset})"
+
+
+def test_closest_point_fit_constrainmin_Calibration_Error() -> None:
```
ReinderVosDeWael, line 147: no capital letters in function names, yes even if they refer to a class ;) (c.f. https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations)

tests/unit/test_calibration.py
```diff
@@ -123,27 +123,80 @@ def test_closest_point_fit() -> None:
...
+    calibrator = calibration.ConstraintedMinimizationCalibration(
+        min_calibration_error=0, max_iterations=1
+    )
+
+    with pytest.raises(calibration.CalibrationError):
+        calibrator.run_calibration(dummy_measure)
+
+
+def test_calibrate_calibration_error() -> None:
+    """Test error when calibration was not possible."""
```
ReinderVosDeWael, line 171: Does this need the full set-up when you're only testing for a CalibrationError?

tests/unit/test_calibration.py
```diff
@@ -226,11 +279,44 @@ def test_run_calibration() -> None:
...
+        result, models.Measurement
+    ), f"was expecting type models.Measurement, object is of type {type(result)}"
+    assert np.allclose(
+        result.measurements, expected_data.measurements, atol=1e-3
+    ), "Measurement data did not match"
+    assert result.time.equals(expected_data.time), "Time series are not equal"
+
+
+def test_run_constrainedmincalibration() -> None:
+    """Testing run fcalibration fuinction for the constrained minimization."""
```
ReinderVosDeWael, line 289: couple typos

ReinderVosDeWael (changes requested):
Overall very minor comments; the only comment of substances pertains to some restructuring of run_calibration in our own calibrator

src/wristpy/processing/calibration.py
```diff
@@ -141,7 +161,7 @@ def __init__(
         self.error_tolerance = error_tolerance
         self.min_calibration_error = min_calibration_error
```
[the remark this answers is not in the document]
Asanto32: Yes, this is wrong throughout. I guess we check if we reach the minimum calibration error. As opposed to 'this is the maximum acceptable calibration error'.?

tests/unit/test_calibration.py
Asanto32: So many typos in this whole PR, sorry. Multitasking got the best of me.

src/wristpy/processing/calibration.py
```diff
@@ -462,3 +439,192 @@ def _get_sampling_rate(timestamps: pl.Series) -> int:
         )
 
         return round(sampling_rate)
+
+
+class ConstraintedMinimizationCalibration(AbstractCalibrator):
```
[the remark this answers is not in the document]
Asanto32: Typo here too ugh...

src/wristpy/processing/calibration.py
```diff
@@ -68,8 +70,26 @@ class LinearTransformation:
     offset: np.ndarray
 
 
-class Calibration:
-    """Implements calibration on accelerometer data, based off of GGIR's implementation.
+class AbstractCalibrator(abc.ABC):
+    """Abstract class defining the interface for the different calibration methods."""
+
+    @abc.abstractmethod
+    def __init__(self) -> None:
```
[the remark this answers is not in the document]
Asanto32: I was thinking about it but couldn't decide on which are mandatory so I left it empty for now

tests/unit/test_calibration.py
Asanto32: I don't think I touched this test, but I based the tests for the new class as clones of the existing GGIR ones. If you mean the function above, I was trying to get the second CalibrationError that we try to Raise explicitly in `run_calibration` (this is actually the line that doesn't seem to be covered) but it might also change based on your comments about organizing our `run_calibration`

src/wristpy/processing/calibration.py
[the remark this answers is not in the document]
ReinderVosDeWael: yeah, pro-tip if you use vscode: put your cursor on a variable, F2, and you can modify its occurence everywhere except docstrings

tests/unit/test_calibration.py
ReinderVosDeWael: it shows as modified in the PR :s, I was referring to test_calibrate_calibration_error

src/wristpy/processing/calibration.py
```diff
@@ -462,3 +439,192 @@ def _get_sampling_rate(timestamps: pl.Series) -> int:
...
+        ) + linear_transformation.offset
+
+        cal_error_initial = np.round(
+            np.mean(abs(np.linalg.norm(no_motion_data, axis=1) - 1)), decimals=5
+        )
+        cal_error_end = np.around(
+            np.mean(abs(np.linalg.norm(no_motion_calibrated, axis=1) - 1)), decimals=5
+        )
+
+        if cal_error_end >= self.min_calibration_error:
```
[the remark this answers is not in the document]
Asanto32: I actually don't see the error for the solution: https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html#scipy.optimize.OptimizeResult

but yes, I can move it into the `fit` function

tests/unit/test_calibration.py
Asanto32: The only thing I changed in that one is the naming for the new `GGIRCalibration` class. I believe the `CalibrationError` only gets raised in `calibrator._calibrate` so it needs the full `dummy_measure`

src/wristpy/processing/calibration.py
```diff
@@ -462,3 +439,191 @@ def _get_sampling_rate(timestamps: pl.Series) -> int:
...
+        scale_bounds = [(0.1, None)] * 3
+        offset_bounds = [(-0.5 * data_range, 0.5 * data_range)] * 3
+        bounds = scale_bounds + offset_bounds
+        result = optimize.minimize(
+            get_distance_to_unit_sphere,
+            initial_guess,
+            options={"maxiter": self.max_iterations},
+            bounds=bounds,
+        )
+        cal_error_end = np.sqrt(result.fun / len(no_motion_data))
```
ReinderVosDeWael, line 559: I _think_ this will error if the optimization fails because result.fun might not exist. If that's correct, or if you just don't feel like confirming, bring this into the result.success if-statement.

ReinderVosDeWael (approved):
One minor thing but otherwie good to go

src/wristpy/processing/calibration.py
Asanto32: I moved it inside the if statement, since it would only be used if result.success == true. Based on my quick reading of the documentation result.fun would still exist even if success == false (it would be the evaluation at the last iteration)

PR #105, opened 2024-09-17
Feature: Adding support for single threshold non-wear detection across watch platforms

### LINKED ISSUE #99: Task: Modify the non-wear detection algorithm to account for various watch types
### Description

There seems to be a significant difference in the range of acceleration values between Actigraph watches and the GENEActiv watches, this difference is enough to change the threshold value for non-wear detection. This task is focused on addressing a few key issues, namely: should/can only the `std_criteria` be used, deciding how and when to set unique criteria thresholds for non-wear detection based on watch type.
[Comparison of Nonwear.pdf](https://github.com/user-attachments/files/16981014/Comparison.of.Nonwear.pdf)

### Tasks

- [x] Modify `detect_nonwear` to default the `range_criteria` to `None` to only use the `std_criteria` for non wear detection. This allows `detect_nonwear` function to work on both .gt3x and .bin files by default without any need for the user to change `range_criteria`

### Freeform Notes

_No response_

Asanto32:
Resolves #99 

Re-based to #71 , this PR modifies the `detect_nonwear` function to default to using a single threshold `std_criteria`. This allows the non-wear detection algorithm to work with both .gt3x and .bin files without the need of chaning the `range_criteria`


Asanto32:
Probably good for a review but might be best to finalize it after the orchestrator is merged? Or maybe just get some feedback on using config.Settings to pass Optional `range_criteria`.

tests/smoke/test_orchestrator_smoke.py
```diff
@@ -40,3 +40,26 @@ def test_orchestrator_different_epoch(
...
+
+@pytest.mark.parametrize(
+    "file_name", [pathlib.Path("test_output.csv"), pathlib.Path("test_output.parquet")]
+)
+def test_happy_path_range_criteria(
+    file_name: pathlib.Path, tmp_path: pathlib.Path, sample_data_gt3x: pathlib.Path
+) -> None:
+    """Happy path for orchestrator."""
+    settings_range_criteria = config.Settings(RANGE_CRITERIA=0.1)
+    results = orchestrator.run(
```
frey-perez, line 53: probably needs a space here to separate arrange and act.

tests/smoke/test_orchestrator_smoke.py
```diff
@@ -40,3 +40,26 @@ def test_orchestrator_different_epoch(
...
+        settings=settings_range_criteria,
+    )
+
+    assert (tmp_path / file_name).exists()
+    assert isinstance(results, orchestrator.Results)
+    assert results.enmo is not None
+    assert results.anglez is not None
+    assert results.nonwear_epoch is not None
+    assert results.sleep_windows_epoch is not None
+    assert results.physical_activity_levels is not None
```
frey-perez, line 65: Just realized there was one change I forgot to make in my PR here, we decided that these assert checks should actually just directly check that they're models.Measurement.

tests/unit/test_metrics.py
```diff
@@ -153,6 +153,32 @@ def test_group_acceleration_data_by_time() -> None:
 )
 def test_compute_nonwear_value_per_axis(
     create_acceleration: pl.DataFrame, modifier: int, expected_result: int
+) -> None:
+    """Test the nonwear value per axis function."""
+    std_criteria = modifier
```
frey-perez, line 158: Not sure this is a necessary alias, perhaps get rid of this and then just change the function call to `metrics._compute_nonwear_value_per_axis(
        acceleration_grouped["X"], std_criteria= modifier
    )`

frey-perez (commented):
Minor comments, nothing too big.

src/wristpy/core/config.py
```diff
@@ -11,6 +12,7 @@ class Settings(pydantic_settings.BaseSettings):
     LIGHT_THRESHOLD: float = 0.03
     MODERATE_THRESHOLD: float = 0.1
     VIGOROUS_THRESHOLD: float = 0.3
+    RANGE_CRITERIA: Optional[float] = None
```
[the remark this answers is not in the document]
Asanto32: As per discussion will remove range_criteria everywhere and re-add if requested.

ReinderVosDeWael (approved):
Y'all are making the PRs too easy on me today.

PR #116, opened 2024-10-06
Refactor the `GgirCalibration` class to `GGIRCalibration`

brutusyhy:
Since GGIR is always referred to in the upper case throughout the documentation, I think it's better to refactor the name to have the same casing as `GGIRSleepDetection` class.


Asanto32:
Hi,
Thanks for the comment as per the style guide we are following: https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations
Class naming should be `UpperCase`.
(Note we will make the change in `GGIRSleepDetection` to `GgirSleepDetection` for consistency

PR #117, opened 2024-10-07
114 task removing settings class

### LINKED ISSUE #114: Task: Removing Settings class
### Description

A bug was discovered in the get_logger function in configs.py with regards to the Settings class. When the logger is created with get_logger, it uses a new settings object, making it impossible to properly log things. The decision has been made to remove the settings class, and set the logging level with the setLogging method. 

The passing of the physical activity levels will also need to be changed to reflect the removal of the Settings class. Logger verbosity arguments should also be added to the cli and orchestrator. 

### Tasks

- [ ] Remove Settings class.
- [ ] Adjust get_logger function.
- [ ] Add arguments to allow the user to set the level through the cli and orchestrator.run() function.
  - CLI verbosity levels should be warning, info, debug. 
- [ ] Adjust the passing of physical activity thresholds to reflect the removal of the Settings class.
  -  Add validation of threshold values again, as the model_validation will no longer be present. 

### Freeform Notes

_No response_

frey-perez:
Resolves #114, Settings class has been removed and the logger has been adjusted accordingly. Additionally a verbosity argument has been added to the CLI, and validation for threshold arguments has been added to compensate for losing the Settings class validation.

### DIFF
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -1,13 +1,19 @@
 """Test logging in config.py."""
 
+import logging
+
 import pytest
 
 from wristpy.core import config
 
 
 def test_get_logger(caplog: pytest.LogCaptureFixture) -> None:
-    """Test the wristpy logger with level set to 20 (info)."""
-    logger = config.get_logger()
+    """Test the wristpy logger with level set to default 20 (info)."""
+    if logging.getLogger("wristpy").handlers:
+        logging.getLogger("wristpy").handlers.clear()
+        logger = config.get_logger()
+    else:
+        logger = config.get_logger()
 
     logger.debug("Debug message here.")
     logger.info("Info message here.")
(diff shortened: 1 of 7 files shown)


tests/unit/test_config.py
```diff
@@ -1,13 +1,19 @@
...
 import pytest
 
 from wristpy.core import config
 
 
 def test_get_logger(caplog: pytest.LogCaptureFixture) -> None:
-    """Test the wristpy logger with level set to 20 (info)."""
-    logger = config.get_logger()
+    """Test the wristpy logger with level set to default 20 (info)."""
+    if logging.getLogger("wristpy").handlers:
```
ReinderVosDeWael, line 12: You can move `logger = ...` out of the if-statement and drop the else statement entirely.

ReinderVosDeWael (approved):
Just one minor comment - no need to ping me again before merging.

PR #118, opened 2024-10-11
fix: remove the requirement for 'time' Series in Measurement to have name 'time'. 

ReinderVosDeWael:
In the current implementation, moving_median works only if the polars Series containing the timepoints already has the "time" name. Since this is not enforced by the Measurement class, it can lead to an error when sorting by time.

This PR updates the moving_median function to ensure that the time column name is always set correctly.

### DIFF
--- a/tests/unit/test_computations.py
+++ b/tests/unit/test_computations.py
@@ -175,7 +175,7 @@ def test_moving_median(window_size: int, expected_output: np.ndarray) -> None:
         measurements=test_matrix, time=dummy_datetime_pl
     )
 
-    test_result = computations.moving_median(test_measurement, window_size=window_size)
+    test_result = computations.moving_median(test_measurement, epoch_length=window_size)
 
     assert test_result.measurements.shape == expected_output.shape, (
         f"measurements array are not the same shape. Expected {expected_output.shape}, "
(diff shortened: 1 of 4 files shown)


Asanto32:
Is this change also needed in `moving_mean` and `moving_std`?
They have the same assumption and do the same pl.concat step

ReinderVosDeWael:
@Asanto32 I've refactored the `computations.py` module to centralize the handling of these things to ensure that bugs like this only need to be fixed in one place. To that end the following changes were made:

- Addition of a `_moving` function that handles the internal moving/rolling functionality.
    - All `moving_` functions now point to this one, in the long run there's an open question on whether the `moving_` functions are necessary, but for this PR I opted to adhere to the existing structure.
- I've centralized the measurement to LazyFrame functionality and DataFrame to Measurement functionality as a function/constructor on the Measurement class.
    - I haven't gone through the codebase to see where this could be further implemented.

tests/unit/test_computations.py
```diff
@@ -162,7 +162,7 @@ def test_moving_median(window_size: int, expected_output: np.ndarray) -> None:
     """Testing proper function of moving median function."""
     dummy_date = datetime(2024, 5, 2)
     dummy_datetime_list = [dummy_date + timedelta(seconds=i) for i in range(3)]
-    dummy_datetime_pl = pl.Series("time", dummy_datetime_list)
+    dummy_datetime_pl = pl.Series(dummy_datetime_list)
```
Asanto32, line 165: would anything bad happen if it was assigned the name "time" here? just redundant since the moving_median enforces it

Asanto32 (approved):

tests/unit/test_computations.py
ReinderVosDeWael: It'd just be ignored

PR #119, opened 2024-10-11
anglez calculation should be arcsin

tigretigre:
As evidenced by the unit tests, the angle relative to the horizontal plan requires arcsin, not arctan.


Asanto32:
As discussed in the issue this is incorrect.

PR #142, opened 2024-11-14
133 directory processing

### LINKED ISSUE #133: Task: Support for full directory processing
### Description

Allow for full directory processing of wristpy pipeline. Users should have the option to provide a directory as input and an output directory, and we will process all files in the input.

### Tasks

- [ ] Create a batch processor to process all accelerometer files in a directory
- [ ] How to handle a directory with mixed inputs (.gt3x and .bin)?

### Freeform Notes

Sample version for .gt3x files that I already have

```
def main() -> None:
    """Main function for batch processing."""
    file_path = Path("/app/data/")
    output_dir = "/app/output/"
    file_names = sorted([file.stem for file in file_path.glob("*.gt3x")])
    for file in file_names:
        input_path = file_path + file + ".gt3x"
        output_path = output_dir + file + "_output/"
        print(f"Processing {input_path}")
        processor.process_file(input_path, output_path)


if __name__ == "__main__":
    main()
```

frey-perez:
Resolves #133. Allows for the run function to be used on single files or directories. The former run() function is now run_file() and run() now calls run_file once in the case of single files or for every valid file in the case of directory inputs. 
The new run() function will still return individual Results objects when returning one file, and a dictionary of Results objects when running entire directories.

### DIFF
--- a/src/wristpy/core/orchestrator.py
+++ b/src/wristpy/core/orchestrator.py
@@ -121,7 +121,7 @@ def run(
     if output is not None:
         output = pathlib.Path(output)
         if not output.is_dir():
-            raise ValueError(f"Output:{output} is not a directory.")
+            raise ValueError("Output is not a valid directory.")
 
     file_names = [
         file for file in itertools.chain(input.glob("*.gt3x"), input.glob("*.bin"))
--- a/src/wristpy/core/orchestrator.py
+++ b/src/wristpy/core/orchestrator.py
@@ -162,9 +162,6 @@ def run(
             verbosity=verbosity,
         )
 
-    if output is None and output_filetype is not None:
-        raise ValueError("If no output is given, output_filetype must be None.")
-
     return _run_directory(
         input=input,
         output=output,
@@ -224,6 +221,9 @@ def _run_directory(
         output from wrist- and hip-worn monitors. Medicine and Science in Sports and
         Exercise, 46(9), 1816-1824.
     """
+    if output is None and output_filetype is not None:
+        raise ValueError("If no output is given, output_filetype must be None.")
+
     if output is not None:
         if output.is_file():
             raise ValueError(
@@ -243,7 +243,7 @@ def _run_directory(
     results_dict = {}
     for file in file_names:
         output_file_path = (
-            output / pathlib.Path(file.stem).with_suffix(output_filetype)  # type: ignore[arg-type]
+            output / pathlib.Path(file.stem).with_suffix(output_filetype)  # type: ignore[arg-type] # if output is defined, so is output_filetype
             if output
             else None
         )
--- a/src/wristpy/core/orchestrator.py
+++ b/src/wristpy/core/orchestrator.py
@@ -187,7 +187,7 @@ def _run_directory(
     epoch_length: Union[int, None] = 5,
     verbosity: int = logging.WARNING,
     output_filetype: Optional[Literal[".csv", ".parquet"]] = None,
-) -> Union[models.OrchestratorResults, Dict[str, models.OrchestratorResults]]:
+) -> Dict[str, models.OrchestratorResults]:
     """Runs main processing steps for wristpy on  directories.
 
     The run_directory() function will execute the run_file() function on entire
@@ -210,8 +210,8 @@ def _run_directory(
         output_filetype: Specifies the data format for the save files.
 
     Returns:
-        All calculated data in a save ready format as a Results object or as a
-        dictionary of OrchestratorResults objects.
+        All calculated data in a save ready format as a dictionary of
+        OrchestratorResults objects.
 
     Raises:
         ValueError: The output given is not a directory.
--- a/src/wristpy/core/orchestrator.py
+++ b/src/wristpy/core/orchestrator.py
@@ -100,7 +100,7 @@ def run(
     ] = "gradient",
     epoch_length: Union[int, None] = 5,
     verbosity: int = logging.WARNING,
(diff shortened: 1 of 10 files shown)


src/wristpy/core/orchestrator.py
```diff
@@ -159,12 +99,119 @@ def run(
...
+    ] = "gradient",
+    epoch_length: Union[int, None] = 5,
+    verbosity: int = logging.WARNING,
+) -> models.Results:
     """Runs main processing steps for wristpy and returns data for analysis.
 
-    The run() function will provide the user with enmo, anglez, physical activity levels
-    sleep detection and nonwear data. All measures will be in the same temporal
+    The run_file() function will provide the user with enmo, anglez, physical activity
+    levels sleep detection and nonwear data. All measures will be in the same temporal
```
Asanto32, line 214:
```suggestion
    levels, detected sleep periods, and nonwear data. All measures will be in the same temporal
```

ReinderVosDeWael (changes requested):

ReinderVosDeWael:
Also update the tutorial

src/wristpy/core/orchestrator.py
```diff
@@ -159,12 +99,119 @@ def run(
...
+            input=input,
+            output=output,
+            thresholds=thresholds,
+            calibrator=calibrator,
+            epoch_length=epoch_length,
+            verbosity=verbosity,
+        )
+    if output is not None:
+        output = pathlib.Path(output)
+        if not output.is_dir():
```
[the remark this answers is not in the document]
frey-perez: Good point, the main point of this was actually to check for user error incase they used a path to a file by mistake. Is it reasonable to just make the check here if output.is_file(): raise error ?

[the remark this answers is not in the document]
ReinderVosDeWael: Yeah seems reasonable.

src/wristpy/core/orchestrator.py
```diff
@@ -159,14 +99,132 @@ def run(
         Literal["ggir", "gradient"],
     ] = "gradient",
     epoch_length: Union[int, None] = 5,
-    verbosity: int = 30,
-) -> Results:
+    verbosity: int = logging.WARNING,
```
ReinderVosDeWael, line 102: I don't see a None return. If that's correct then no need for an Optional return.

src/wristpy/core/orchestrator.py
```diff
@@ -159,14 +99,132 @@ def run(
...
+            verbosity=verbosity,
+        )
+    if output is not None:
+        output = pathlib.Path(output)
+        if output.is_file():
+            raise ValueError(
+                "Output is a file, but must be a directory when input is a directory."
+            )
+
+    if output_filetype is None or output_filetype not in VALID_FILE_TYPES:
```
ReinderVosDeWael, line 172: Couldn't this just be
 
 `if output_filetype not in VALID_FILE_TYPES` since None is not in VALID_FILE_TYPES?

ReinderVosDeWael: This errors if output is None and output_filetype is None, but this should be fine right?

src/wristpy/core/orchestrator.py
```diff
@@ -159,14 +99,132 @@ def run(
...
+            else None
+        )
+        logger.debug(
+            "Processing directory: %s, current file: %s, save path: %s",
+            input,
+            file,
+            output_file_path,
+        )
+        try:
+            results_dict[file.stem] = run_file(
```
ReinderVosDeWael, line 197: Include the suffix in the key; this will overwrite when you have e.g. `subject1.bin` and `subject1.gt3x` in the same directory.

tests/unit/test_orchestrator.py
```diff
@@ -106,12 +108,79 @@ def test_save_results(
...
+    ):
+        orchestrator.run(input=input_dir, output=bad_output_dir)
+
+
+@pytest.mark.parametrize("invalid_file_type", [".zip", None])
+def test_bad_file_type(
+    tmp_path: pathlib.Path, invalid_file_type: Optional[str]
+) -> None:
+    """Test run function when output file type is invalid."""
+    VALID_FILE_TYPES = (".csv", ".parquet")
```
ReinderVosDeWael, line 174: import this from the module i.e. `orchestrator.VALID_FILE_TYPES`

ReinderVosDeWael (changes requested):
couple of minor adjustments

src/wristpy/core/orchestrator.py
frey-perez: Good point, I'll just tab it over once so that it's within the output is not None check

frey-perez: To your first point I think i did that to appease a mypy error a little further down when I use with_suffix (not compatible with None). I'll just add another ignore there.

src/wristpy/core/orchestrator.py
```diff
@@ -159,14 +99,192 @@ def run(
...
+    file_names = list(itertools.chain(input.glob("*.gt3x"), input.glob("*.bin")))
+
+    if not file_names:
+        raise FileNotFoundError(f"Directory {input} contains no .gt3x or .bin files.")
+
+    results_dict = {}
+    for file in file_names:
+        output_file_path = (
+            output / pathlib.Path(file.stem).with_suffix(output_filetype)  # type: ignore[arg-type]
+            if output
```
ReinderVosDeWael, line 247: # If output is defined, so is output_filetype.

ReinderVosDeWael (approved):

PR #143, opened 2024-11-20
Feat/issue 102/handle idle sleep mode

Asanto32:
Support to handle the `idle_sleep_mode` for Actigraph watches. This function fills in time gaps in the data created by enabling `idle_sleep_mode`. In cases when the samples are not linearly spaced within one second, the data is resampled.


Asanto32:
This is ready for review now. Small chance of merge conflicts with the orchestrator compared to the new one with full directory processing but I tried to make sure it should work together.

Note: when there are significant amounts of `idle_sleep_mode` gaps it does slow down processing. Not due to the `idle_sleep_mode_imputation` (which only takes 5-10s), but from the addition of significantly more data points (example test case on my computer adds in more than 50% of the data thus increase in processing time by ~50%)

ReinderVosDeWael (approved):

ReinderVosDeWael (changes requested):
Assess interaction with all locations where imputed data will be used e.g. calibration

ReinderVosDeWael (approved):

PR #151, opened 2024-12-17
Mean Amplitude Deviation metric

### LINKED ISSUE #132: Task: Implement the Mean Amplitude Deviation in `metrics.py`
### Description

Mean Amplitude Deviations can be used as a metric to determine physical activity levels, and has been shown to correlate well with HR in adolescents. The resulting metric is in _g_ force, and the reference paper provides some suggested threshold cut points to categorize activity levels based on this metric ( "the MAD cut-points were 29 and 27 (light), 338 and 330 (moderate), and 604 and 558 (vigorous)."). A promising attribute of this metric is that it is device agnostic, and suggested cut points can be applied across different accelerometers. 

This metric is calculated on the raw accelerometer data. The function can take the acceleration `Measurement` object as input and return a new `Measurement` object that contains the MAD values. 

Data was analysed in 5s epochs. where MAD can be defined as the average distance of data points to the mean in a given epoch:
MAD = (1/n) sum{abs(r<sub>i</sub> - r<sub>u</sub>)}, where r here represents the _g_ -force (Euclidean norm of the raw acceleration data), and r<sub>i</sub> is the ith resultant sample within the epoch of interest, r<sub>u</sub> the mean resultant value of the epoch of interest, n the number of samples in the epoch.

References:
https://bmcsportsscimedrehabil.biomedcentral.com/articles/10.1186/s13102-015-0010-0
https://pubmed.ncbi.nlm.nih.gov/24393233/

### Tasks

- [x] Compute the MAD values based on the norm of the raw acceleration values
- [ ] Add to CLI and python entry point (orchestrator)

### Freeform Notes

_No response_

Asanto32:
Resolves #132 

Adding the mean amplitude deviation metric, alternative to enmo.


Asanto32:
Might wait to merge the bugfix PR first

frey-perez (commented):
Don't have much to add here except adding a little more testing for non-zero expected values. Don't know how necessary that is, given that the function is relatively simple.

ReinderVosDeWael (commented):
How do I use this?

src/wristpy/processing/metrics.py
```diff
@@ -58,6 +58,44 @@ def angle_relative_to_horizontal(
     return models.Measurement(measurements=angle_degrees, time=acceleration.time)
 
 
+def mean_amplitude_deviation(acceleration: models.Measurement) -> models.Measurement:
+    """Calculate the mean amplitude deviation of the acceleration data.
+
+    An alternative to ENMO to quantify the intensity of physical activity.
+    It is calculated as the mean of the absolute difference between acceleration values
+    and the mean of those acceleration values in a 5s window.
```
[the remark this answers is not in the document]
Asanto32: The paper explicitly said 5s but it can be a variable that defaults to 5s?

Asanto32:
Ready for re-review @ReinderVosDeWael  @frey-perez 

Changed how the thresholds argument was being handled and moved some of the error handling around. Let me know what you think. I was having trouble with the None type or List of floats using nargs = 3 (I tried `nargs= '?' ` but no luck), hence the new `_none_or_float_list` function.

src/wristpy/core/cli.py
```diff
@@ -53,18 +53,24 @@ def parse_arguments(args: Optional[List[str]] = None) -> argparse.Namespace:
         help="Pick which calibrator to use. Can be 'ggir' or 'gradient'.",
     )
 
+    parser.add_argument(
+        "-a",
+        "--activity_metric",
+        type=str,
```
ReinderVosDeWael, line 59: IIRC you can set this to `lambda s: s.lower()` to accept any case but force it to always be lower case.

src/wristpy/core/orchestrator.py
```diff
@@ -98,12 +98,13 @@ def format_nonwear_data(
...
     input: Union[pathlib.Path, str],
     output: Optional[Union[pathlib.Path, str]] = None,
-    thresholds: Tuple[float, float, float] = (0.0563, 0.1916, 0.6958),
+    thresholds: Optional[Tuple[float, float, float]] = None,
     calibrator: Union[
         None,
         Literal["ggir", "gradient"],
     ] = "gradient",
     epoch_length: Union[int, None] = 5,
+    activity_metric: Literal["ENMO", "MAD"] = "ENMO",
```
ReinderVosDeWael, line 107: `ggir` in the other argument is all lower case. Pick either all upper or all lower for acronyms, don't mix the two.

src/wristpy/processing/metrics.py
```diff
@@ -58,6 +58,53 @@ def angle_relative_to_horizontal(
...
+        doi: 10.1111/cpf.12127. Epub 2014 Jan 7. PMID: 24393233.
+    """
+    acceleration_magnitude = np.linalg.norm(acceleration.measurements, axis=1)
+
+    mad_lf = pl.LazyFrame(
+        {"time": acceleration.time, "acceleration_magnitude": acceleration_magnitude}
+    ).set_sorted("time")
+
+    mad_df = (
+        mad_lf.group_by_dynamic(index_column="time", every=f"{epoch_length}s")
```
ReinderVosDeWael, line 90: Could convert to nanoseconds and set epoch_length to a float? That'd allow for epoch lengths of e.g. 3.5seconds

src/wristpy/core/cli.py
```diff
@@ -122,21 +131,52 @@ def main(
...
+        thresholds=None
+        if arguments.thresholds is None
+        else cast(Tuple[float, float, float], tuple(arguments.thresholds)),
         epoch_length=None if arguments.epoch_length == 0 else arguments.epoch_length,
         verbosity=log_level,
         output_filetype=arguments.output_filetype,
     )
+
+
+def _none_or_float_list(value: str) -> Optional[List[float]]:
```
ReinderVosDeWael, line 150: Yeah.. I'm starting to like the idea of checking out tyro. There's nothing wrong with this function, but this would be so much easier if we can just define a Pydantic model.

src/wristpy/core/cli.py
Asanto32: Should I make this change for the calibrator as well? Note I have switched the activity metric choices to lower case,  "enmo" and "mad"

src/wristpy/core/orchestrator.py
Asanto32: Switched to all lower case, this goes for the argparse options in the CLI as well

src/wristpy/core/cli.py
Asanto32: Potentially a new issue to be added to our task list

src/wristpy/core/cli.py
Asanto32: Made the change in both cases to handle any mixed case entries/typos in user input

ReinderVosDeWael (approved):

Issue #127, opened 2024-10-24
Task: Implement Monitor Independent Monitor Summary unit (MIMS)


frey-perez:
### Description

In response to reviewer feedback it's become clear that there is strong interest in acceleration metrics beyond the basic ENMO metric we currently report. One such metric of interest is Monitor-Independent Movement Summary (MIMS). MIMS units provide an open-source, universal metric for summarizing acceleration data across different devices, eliminating inconsistencies caused by device-specific factors. This standardization enables better cross-study comparisons and supports the integration of both research and consumer-grade devices. Furthermore MIMS has been shown to capture information just above the devices noise threshold, making it more sensitive to subtle movements (as compared to other metrics like Activity Counts). 

Wristpy will add a function to the metrics module that will calculate MIMS. Adding MIMS to the Wristpy orchestrator module, and data output will be completed in a future pull request. 

[Source for algorithm implementation:](https://pmc.ncbi.nlm.nih.gov/articles/PMC8301210/) 

### Tasks

- [x] Interpolation

  - Natural cubic spline interpolator to resample to 100 Hz. 

- [ ] Extrapolation

  1. Find "Maxed- out" samples
    - Use mathematical model depicted below to find the probabilty that a sample at n is "maxed-out"
    - P<sub>n</sub> (E|x = |d| − d<sub>0</sub>) = F( x = |d| − d<sub>0</sub>; k, θ) = γ(k, θx)   / Γ(k) if x < 0, P<sub>n</sub> (E|x) = 0 
  2. Find "Maxed-out" edges
    - The maxed out regions are determined from the difference between confidence probabilities of adjacent samples. (Q)
    - Q(n − 1, n) = P<sub>n</sub> − P<sub>n</sub>−1
    - If Q is greater than some threshold C (set at 0.5), that sample is a boundary. The sign determines if it is a hill (+) or a valley (-).
    - Samples within T seconds outside of the left and right boundary are selected for the signal reconstruction
  3. Extend "maxed-out" signal to make it smoother (as though it had not hit the limits of the devices range)
    - Values within T seconds are modeled using weighted spline regression, where weights are set as probability of not being maxed out. The peak is estimated by finding where the two regression curves meet, and the maxed-out data points are replaced with smoothly fitted values to keep the sampling intervals even.

- [ ] Bandpass Filter
[...]

PR #158, opened 2025-01-17
Mims interpolate rebase

frey-perez:
Implements Checkbox 1 of task https://github.com/childmindresearch/wristpy/issues/127. The interpolate function is the first step in the mims algorithm. It uses a natural cubic spline to interpolate the data to 100 hz.


tests/unit/test_metrics.py
```diff
@@ -226,3 +228,55 @@ def test_detect_nonwear(
...
+
+    interpolated_acceleration = metrics.interpolate_measure(
+        acceleration=acceleration, new_frequency=100
+    )
+    interpolated_ms = interpolated_acceleration.time.dt.epoch(time_unit="ms").to_numpy()
+
+    assert len(expected_time) == len(
+        interpolated_acceleration.time
+    ), "Timestamp series are not the same length."
+    assert np.allclose(
```
Asanto32, line 254: Shouldn't the timestamps be exact? We are interpolating the time to 100 Hz in the same range? It is only the "y-values/data" that could be slightly off?

frey-perez: Ah yes, so this is a relic of how I had things set up earlier. Our data is at ns resolution, and the R package will only do μs, so they weren't exactly the same. I changed our resolution in the function to μs, but forgot to change this. 

So i can either use nanoseconds in our implementation, and use all close in the test, or keep using μs and make sure they're exactly the same. Not sure which is preferable.

ReinderVosDeWael: Can't you just test at microsecond precision but leave our implementation as is?

ReinderVosDeWael: yeah id just leave ours as is and account for the inaccuracy in testing.

src/wristpy/processing/metrics.py
```diff
@@ -251,27 +251,35 @@ def _cleanup_isolated_ones_nonwear_value(nonwear_value_array: np.ndarray) -> np.
...
-            epoch_time_seconds, acceleration.measurements[:, axis]
+            epoch_time_seconds, acceleration.measurements[:, axis], bc_type="natural"
         )
         interpolated_data[:, axis] = cubic_spline(interpolated_time)
 
-    new_time_ns = (interpolated_time * 1e9).astype(np.int64)
+    new_time_ns = (interpolated_time * 1e6).astype(np.int64)
     new_time_series = pl.Series(
-        "interpolated_time", new_time_ns, dtype=pl.Datetime("ns")
+        "interpolated_time", new_time_ns, dtype=pl.Datetime("us")
```
ReinderVosDeWael, line 282: The variable name is ns, the dtype is us?

tests/unit/test_metrics.py
```diff
@@ -226,3 +228,55 @@ def test_detect_nonwear(
...
+    test_data = readers.read_watch_data(sample_data_gt3x)
+
+    interpolated_acceleration = metrics.interpolate_measure(
+        acceleration=test_data.acceleration, new_frequency=100
+    )
+
+    assert (
+        expected_acceleration.shape == interpolated_acceleration.measurements.shape
+    ), "Shape error."
+    for axis in range(3):
```
ReinderVosDeWael, line 275: For simplicity, could this not just be a `np.allclose()`

ReinderVosDeWael (changes requested):
One thing I'm still missing is some explanation of where the file `actigraph_interpolation*` came from.

tests/unit/test_metrics.py
frey-perez: Yea the issue was that np.allclose wouldn't work for any reasonable atol or rtol because there are a small handful of outliers that make it fail, so the correlation was the work around. Should I leave as is or deal with it another way?

ReinderVosDeWael (approved):

PR #161, opened 2025-01-31
Feature/issue 130/implement agcounts

Asanto32:
Implement the Actigraph activity count metric


Asanto32:
Only tests null and max possible count value. Modified resample from #103 (modified version will be needed fro temperature non-wear PR)

Asanto32:
Ok think this is good for a look over now.

tests/smoke/test_orchestrator_smoke.py
```diff
@@ -49,6 +49,29 @@ def test_orchestrator_different_epoch(
     assert isinstance(results.physical_activity_levels, models.Measurement)
 
 
+def test_orchestrator_none_epoch_ag_count(
```
ReinderVosDeWael, line 52: Given the ever-growing set of if-statements in the runner, this could do with a spy to assess that an expected function call is actually performed.

tests/unit/test_computations.py
```diff
@@ -184,3 +184,105 @@ def test_moving_median(window_size: int, expected_output: np.ndarray) -> None:
...
+            for secs in range(round(expected_time_length))
+        ]
+    )
+
+    actual = computations.resample(measurement, delta_t)
+    actual_sampling_rate = 1 / (actual.time[1] - actual.time[0]).total_seconds()
+
+    assert len(actual.time) == expected_time_length
+    assert time[0] == actual.time[0]
+    assert round(actual_sampling_rate) == expected_sampling_rate
```
ReinderVosDeWael, line 238: Why does this need to be rounded? Could it do with a `np.isclose()` to ensure we don't have a case like 29.5 == 30?

tests/unit/test_computations.py
```diff
@@ -184,3 +184,105 @@ def test_moving_median(window_size: int, expected_output: np.ndarray) -> None:
...
+    """Test the downsampling with 60 and 30Hz.
+
+    This test is specifically to catch any issues that arise with sampling
+    associated with imprecise detla_t values.
+    This test will fail without the `start_by` option in the
+    polars`group_by_dynamic` function used by computations.resample.
+    """
+    time = [
+        datetime(1990, 1, 1) + timedelta(microseconds=16666 * secs)
+        for secs in range(100)
```
ReinderVosDeWael, line 217: Some confusing variable naming here, since you're multiplying it by 16666 microseconds, its really more of a counter than a 'second'. You could use a more generic name here `idx`, `index`, etc..

tests/unit/test_computations.py
```diff
@@ -184,3 +184,105 @@ def test_moving_median(window_size: int, expected_output: np.ndarray) -> None:
...
+    assert np.allclose(
+        actual.time.dt.timestamp().to_numpy(), expected_time.dt.timestamp().to_numpy()
+    )
+
+
+def test_resample_upsample() -> None:
+    """Test the upsampling happy-path."""
+    time = [
+        datetime(1990, 1, 1) + timedelta(microseconds=16666 * secs)
+        for secs in range(10)
```
ReinderVosDeWael, line 248: Same variable name comment as prior test

tests/unit/test_metrics.py
```diff
@@ -242,3 +242,49 @@ def test_mean_amplitude_deviation_function(create_acceleration: pl.DataFrame) ->
...
+        {
+            "X": X * 100,
+            "Y": Y * 100,
+            "Z": Z * 100,
+            "time": test_time,
+        }
+    )
+    acceleration = models.Measurement.from_data_frame(acceleration_polars_df)
+
+    expected_result = np.sqrt(3 * ((50 * 128) ** 2))
```
ReinderVosDeWael, line 284: This could do with some explanation. e.g. `# Magnitude of 3D array of 120 (max value) at 50Hz.`

ReinderVosDeWael (changes requested):
A few comments here and there. There's a lot of 'magic numbers' in this PR i.e. numbers for which its not really obvious where they come from or what they are.

Two solutions for the magic number problem: 1) create a named variable for them, or 2) in-line comment (yes, you heard that right! :P ) it with the source/explanation. _Subjectively_, I'd generally favor 1 over 2, but 2 could be preferred in cases where you're referring to e.g. coefficients derived in a paper.

src/wristpy/processing/metrics.py
```diff
@@ -105,6 +106,104 @@ def mean_amplitude_deviation(
...
+    References:
+        [1] A. Neishabouri et al., “Quantification of acceleration as activity counts
+        in ActiGraph wearable,” Sci Rep, vol. 12, no. 1, Art. no. 1, Jul. 2022,
+        doi: 10.1038/s41598-022-16003-x.
+    """
+    logger.debug("Running activty count physical activity metric.")
+
+    epoch_length_nanoseconds = round(epoch_length * 1e9)
+
+    input_coef = np.array(
```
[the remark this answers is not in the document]
Asanto32: Yeah these are all from the reference

src/wristpy/processing/metrics.py
```diff
@@ -105,6 +106,104 @@ def mean_amplitude_deviation(
...
+        output_coef,
+        acceleration_30Hz.measurements,
+        zi=np.repeat(
+            initial_conditions, acceleration_30Hz.measurements.shape[1], axis=-1
+        )
+        * acceleration_30Hz.measurements[0],
+        axis=0,
+    )
+
+    scaled_acceleration = acceleration_bpf * (3 / 4096) / (2.6 / 256) * 237.5
```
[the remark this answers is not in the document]
Asanto32: yeah itsn the scaling factor from [1], I'll make it a variable

tests/unit/test_computations.py
Asanto32: Yeah isclose() will probably work. It's just that the rounding to nearest nanosecond from (1/30)s gives a samping rate of 30.0003

tests/unit/test_metrics.py
Asanto32: Got it, will add

src/wristpy/core/orchestrator.py
```diff
@@ -388,15 +394,28 @@ def _run_file(
...
     anglez = metrics.angle_relative_to_horizontal(calibrated_acceleration)
     sleep_detector = analytics.GgirSleepDetection(anglez)
     sleep_windows = sleep_detector.run_sleep_detection()
 
     if epoch_length is not None:
-        activity_measurement = computations.moving_mean(
-            activity_measurement, epoch_length=epoch_length
-        )
+        if activity_metric != "ag_count":
+            activity_measurement = computations.moving_mean(
```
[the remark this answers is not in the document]
Asanto32: Actually looking back I think there's a bug that can arise with MAD if epoch_length is `None`. 

The main issue is that `enmo` and `anglez` have the same temporal resolution as acceleration (no downsampling). Whereas `mad` and `ag_count` will always be downsampled (relative to acceleration/sensor sampling rate) to `epoch_length`, so if, for some unknown reason, someone specifically asked for `epoch_length = None` and for `physical_activity_metric = mad/ag_count` I needed to make sure that anglez will get downsampled to the epoch_length used in `mad` and `ag_count`. As `anglez` is one of the outputs of our orchestrator, it needs to be at the same temporal resolution as physical activity.

Basically I'm trying to protect the user from breaking this.


In summary. I will go over it and re-work the logic so that it is clear.

src/wristpy/core/orchestrator.py
```diff
@@ -388,15 +394,28 @@ def _run_file(
...
+                activity_measurement, epoch_length=epoch_length
+            )
         anglez = computations.moving_mean(anglez, epoch_length=epoch_length)
+    elif (epoch_length is None) and (activity_metric == "ag_count"):
+        anglez = computations.moving_mean(anglez, epoch_length=5)
```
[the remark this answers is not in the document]
Asanto32: Correct, forgot to change it after adding ag_counts (and, as mentioned above, it should have been changed when I added `mad`)

Asanto32:
I left the big if-else logic statement I was wondering if we should move to a dictionary/key look up considering we will also add MIMS to this list, another thought we could discuss tomorrow is that we currently only allow one choice for physical activity metric, but in theory we could do any number?

src/wristpy/processing/calibration.py
```diff
@@ -600,3 +600,53 @@ def _extract_no_motion(
...
+        """
+        if name == "ggir":
+            self._calibrator = GgirCalibration()
+        elif name == "gradient":
+            self._calibrator = ConstrainedMinimizationCalibration()
+        else:
+            raise ValueError("Unknown calibrator.")
+
+    def run(
+        self, acceleration: models.Measurement, *, return_input_on_error: bool = False
```
frey-perez, line 627: I know we call it in run_file as True, but shouldn't we just default to True? Additionally we can't actually access or change return_input_on_error from run_file(), run_directory or run().

I'm not sure it's important, but we should change it if this is something we care about the user (who for the most part will never use the CalibrationDispatcher directly) being able to do this from the CLI or orchestrator.

src/wristpy/core/computations.py
```diff
@@ -106,3 +110,62 @@ def moving_median(array: models.Measurement, epoch_length: int) -> models.Measur
...
+        The resampled measurement.
+
+    Raises:
+        ValueError: Raised for zero or negative delta_t.
+    """
+    if delta_t <= 0:
+        msg = "delta_t must be positive."
+        raise ValueError(msg)
+
+    all_delta_t = (measurement.time[1:] - measurement.time[:-1]).unique()
```
frey-perez, line 133: Seems to me like the use of unique here could go awry unexpectedly? Might be safer to check for the mode or median?

Asanto32: Hmmm I think that would work, and then remove accessing `all_delta_t[0]` below. I had left this from when Reinder wrote this function. But so far in all the real examples the weird cases that have arisen are when sampling at 30 or 60Hz, we actually get two delta_t values. The shortest one (usually closest to 1/30 or 1/60, respectively) and then there's a second one that is essentially due to the rounding error. Accessing `all_delta_t[0]` mitigates any issues, but median might work too.

tests/unit/test_calibration.py
```diff
@@ -348,3 +348,12 @@ def test_run_chunked_calibration() -> None:
         result.measurements, expected_data.measurements, atol=1e-3
     ), "Measurement data did not match"
     assert result.time.equals(expected_data.time), "Time series are not equal"
+
+
+def test_calibration_dispatcher_bad() -> None:
```
frey-perez, line 353: I think we should add testing for when the calibrator errors and we want to return the input or raise.

src/wristpy/core/computations.py
Asanto32: Yeah, I like this and everything still works. Will make the change. Can revert back if Reinder has a reason I haven't thought of

src/wristpy/core/orchestrator.py
```diff
@@ -158,6 +164,8 @@ def run(
         thresholds = thresholds or (0.0563, 0.1916, 0.6958)
     elif activity_metric == "mad":
         thresholds = thresholds or (0.029, 0.338, 0.604)
+    elif activity_metric == "ag_count":
+        thresholds = thresholds or (100, 3000, 5200)
```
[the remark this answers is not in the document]
Asanto32: This is only here for the default pipeline, handling the logic if someone forgot to provide threshold values, right? I think if someone is accessing run_file/run_directory directly they have the knowledge/skill to set their thresholds?

src/wristpy/processing/calibration.py
Asanto32: Yeah, again, ha, Reinder added this in but IMO I would always have this True and not even be an option/variable

tests/unit/test_calibration.py
Asanto32: See my other comment about my opinion on if this should even be an option aha, but yeah I can add that easily

src/wristpy/core/cli.py
```diff
@@ -57,7 +57,7 @@ def parse_arguments(args: Optional[List[str]] = None) -> argparse.Namespace:
         "-a",
         "--activity_metric",
         type=lambda s: s.lower(),
-        choices=["enmo", "mad"],
+        choices=["enmo", "mad", "ag_count"],
         default="enmo",
         help="Pick which physical activity metric should be used. "
         "This will be used to determine physical activity categorization. "
```
frey-perez, line 63:
```suggestion
        help="Pick which physical activity metric should be used. "
        "This will be used to determine physical activity categorization. "
        "Can be 'enmo', 'mad', or 'ag_count'."
```

src/wristpy/core/orchestrator.py
```diff
@@ -313,11 +321,13 @@ def _run_file(
...
         thresholds: The cut points for the light, moderate, and vigorous thresholds,
-            given in that order. Values must be asscending, unique, and greater than 0.
+            given in that order. Values must be ascending, unique, and greater than 0.
             Default values are optimized for subjects ages 7-11 [1].
         calibrator: The calibrator to be used on the input data.
         epoch_length: The temporal resolution in seconds, the data will be down sampled
-            to. If None is given no down sampling is preformed.
+            to. If None is given, and `enmo` is the chosen physical activity metric,
+            no down sampling is preformed. Otherwise, for `mad` and `ag_count`, a chosen
+            epoch length of `None` will be forced to the default value of 5.0s.
```
frey-perez, line 330: None will cause an error to be raised right? We should change the last line here. Also the changes in this doc string should be duplicated in run_file and run_dir.

Asanto32: Forgot to change after we modified that.

frey-perez (commented):
I think it's largely in good shape just had a few comments. 

One thing that did occur to me as a small quality of life feature for the user is making it so that the selected metric becomes the name of the physical_activity_metric column. That can happen later if we want to make that change. 

Beyond that I just added some small things to consider and double checking doc strings for consistency after all the changes we've made.

Asanto32:
> I think it's largely in good shape just had a few comments.
> 
> One thing that did occur to me as a small quality of life feature for the user is making it so that the selected metric becomes the name of the physical_activity_metric column. That can happen later if we want to make that change.
> 
> Beyond that I just added some small things to consider and double checking doc strings for consistency after all the changes we've made.

Yeah I was also wondering about allowing multiple physical activity metrics to be output.

And yeah, docstrings are getting long. I think as part of the next release we should do a thorough re-read. Or outsource to someone who isn't so deep in it to look it over aha

Asanto32:
Ok added the test to ensure that error gets raised, I used both `unittest.mock` and `pytest_mock` so not sure if that is correct.

@reindervdw-cmi (can't re-request review)

reindervdw-cmi:
Tell me you used AI to generate your code without telling me you used AI to generate your code ;). 

Yeah prefer using only one framework unless there's something in both frameworks that's you both strictly need that isn't available in the other. I generally use `pytest` so thats what I can help with most

reindervdw-cmi (approved):

PR #175, opened 2025-03-14
Update docs.yaml

### LINKED ISSUE #174: Bug: Failing publish docs workflow
### Description

Failing dependency installation with PEP 517 error (same as ruff test issue). Need to switch to `poetry install --only dev`

### Steps to Reproduce

Run docs.yaml

### Expected Results

Action should not fail.

### Actual Results

Action fails due to PEP 517 build error

### Additional Notes

_No response_

Asanto32:
Resolves #174 
Should fix build docs action bug

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -22,6 +22,8 @@ jobs:
       run: |
         python -m pip install poetry
         poetry install --only dev --with docs
+    - name: Install pdoc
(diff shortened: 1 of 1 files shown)


Asanto32:
Should be good now @reindervdw-cmi (after the war over a space with pre-commit)

reindervdw-cmi (approved):

PR #181, opened 2025-04-09
Mims bandpass

frey-perez:
This pull request addresses the implementation of the butterworth_filter and the aggregation steps of the MIMS algorithm. 

The filtering step applies a butterworth filter to the data, and the aggregation step finds the area under the curve (AUC) for each epoch and returns a value per axis. 

To reference the original implementation please see following highlighted sections. 

- [Application of filter in mims_unit module.](https://github.com/mHealthGroup/MIMSunit/blob/master/R/mims_unit.R#L449-L479)
- [Module where the iir function is defined](https://github.com/mHealthGroup/MIMSunit/blob/master/R/filter.R)
- [Application of aggregation in mims_unit module.](https://github.com/mHealthGroup/MIMSunit/blob/master/R/mims_unit.R#L490-L522)
- [Module where the aggregation function is defined](https://github.com/mHealthGroup/MIMSunit/blob/master/R/aggregate_over_epoch.R)

It should be noted that the aggregate_over_epoch module also contains an additional function: aggregate_for_orientation(). This is an optional, and complimentary output that does not directly affect AUC values, and will be implemented in a separate PR. 

Finally It should also be noted that after per axis values are calculated the mims_unit algorithm gives the user a choice to calculate the total unit using a sum, or vector magnitude[(see here)](https://github.com/mHealthGroup/MIMSunit/blob/master/R/mims_unit.R#L524-L557). This value will handled by the overall mims function and calculated after aggregation.


reindervdw-cmi (approved):

reindervdw-cmi:
Other than the tuple typing comments, feel free to ignore these comments as they're more styling than anything else.

PR #183, opened 2025-04-14
Feature/issue 106/modify sleep windows-refactor sleep output

Asanto32:
Remove the overlaps between sleep and nonwear periods. Re-factor how we convert the sleep windows to a Measurement as part of the output.

### DIFF
--- a/src/wristpy/processing/analytics.py
+++ b/src/wristpy/processing/analytics.py
@@ -301,9 +301,9 @@ def _find_periods(
 
 
 def _fill_false_blocks(boolean_array: np.ndarray, gap_block: int) -> np.ndarray:
-    """Helper function to fill gaps in SPT window that are less than gap_blocks.
+    """Helper function to fill gaps in any window that are less than gap_blocks.
 
-    We find the first non-zero in the sleep_idx_array, if there are none ,
+    We find the first non-zero in the boolean_array, if there are none ,
     we return the initial array.
     We then iterate over the array and count every zero between ones
     (skipping the leading zeros),
@@ -314,7 +314,7 @@ def _fill_false_blocks(boolean_array: np.ndarray, gap_block: int) -> np.ndarray:
         gap_block: the length of the gap that needs to be filled.
 
     Returns:
-        A numpy array with 1s, typically for identified SPT windows.
+        A booelan numpy array where true, typically, idicates the SPT windows.
     """
     n_zeros = 0
     first_one_idx = next(
@@ -417,12 +417,15 @@ def sleep_cleanup(
 
     sleep = _sleep_windows_as_measurement(nonwear_measurement.time, sleep_windows)
 
-    filtered_sleep = sleep.measurements - nonwear_measurement.measurements
+    filtered_sleep = np.logical_and(
+        sleep.measurements,
+        np.logical_not(nonwear_measurement.measurements.astype(bool)),
+    )
     cleaned_sleep = np.logical_not(
         _fill_false_blocks(np.logical_not(filtered_sleep), num_samples_15min)
     )
 
-    return models.Measurement(time=sleep.time, measurements=cleaned_sleep.astype(int))
+    return models.Measurement(time=sleep.time, measurements=cleaned_sleep)
 
 
 def _sleep_windows_as_measurement(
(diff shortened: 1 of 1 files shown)


Asanto32:
@reindervdw-cmi your type of PR

frey-perez (commented):
looks solid to me!

tests/unit/test_analytics.py
```diff
@@ -264,3 +195,30 @@ def test_bad_physical_activity_thresholds() -> None:
...
+            onset=dummy_date + datetime.timedelta(minutes=10),
+            wakeup=dummy_date + datetime.timedelta(minutes=40),
+        )
+    ]
+
+    result = analytics.sleep_cleanup(
+        sleep_windows=sleep_windows, nonwear_measurement=nonwear_measurement
+    )
+    expected_result = np.zeros(len(nonwear_measurement.time))
+    expected_result[600:2000] = 1
```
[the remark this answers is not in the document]
Asanto32: Yeah good catch. Moving up.

src/wristpy/processing/analytics.py
```diff
@@ -440,3 +391,66 @@ def compute_physical_activty_categories(
...
+    Returns:
+        A Measurement instance with the cleaned sleep data.
+    """
+    logger.debug("Starting the sleep Window cleanup.")
+    temporal_resolution = nonwear_measurement.time[1] - nonwear_measurement.time[0]
+    num_samples_15min = int(15 * 60 / temporal_resolution.total_seconds())
+
+    sleep = _sleep_windows_as_measurement(nonwear_measurement.time, sleep_windows)
+
+    filtered_sleep = sleep.measurements - nonwear_measurement.measurements
```
reindervdw-cmi, line 420: These should both be boolean right? If yes, use a logical condition unless `False - True == -1` is intended behavior.

src/wristpy/processing/analytics.py
```diff
@@ -440,3 +391,66 @@ def compute_physical_activty_categories(
...
+    cleaned_sleep = np.logical_not(
+        _fill_false_blocks(np.logical_not(filtered_sleep), num_samples_15min)
+    )
```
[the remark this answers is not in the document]
reindervdw-cmi: I'm fine with this

src/wristpy/processing/analytics.py
```diff
@@ -440,3 +391,66 @@ def compute_physical_activty_categories(
...
+            reference Measurement.
+        sleep_windows: The list of sleep windows, where the entries are
+            instances of the SleepWindow class.
+
+    Returns:
+        A new Measurement instance with the sleep values.
+    """
+    logger.debug("Converting sleep windows to measurement.")
+
+    sleep_value = np.zeros(len(ref_measurement_time), dtype=int)
```
reindervdw-cmi, line 447: Is using integers rather than booleans intentional?

src/wristpy/processing/analytics.py
Asanto32: Huh... this never happened but you are right, current implementation can have a -1 values that wouldn't really make sense. The nonwear measurements isn't a a boolean, it is an array of int 0/1. So I will convert it and then fix the line for sleep_measurement to be boolean not int

src/wristpy/processing/analytics.py
Asanto32: Initially it was to work with the integer output of the nonwear measurement, will switch to boolean.

src/wristpy/processing/analytics.py
```diff
@@ -339,50 +300,40 @@ def _find_periods(
...
     Args:
-        non_wear_array: The non-wear array generated from metrics.detect_nonwear
-        sleep_windows : The list of sleep windows, where the entries are
-            instances of the SleepWindow class.
+        boolean_array: A generic boolean array, typically the SPT window.
+        gap_block: the length of the gap that needs to be filled.
 
     Returns:
-        A List of the filtered sleep windows.
+        A numpy array with 1s, typically for identified SPT windows.
```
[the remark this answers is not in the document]
Asanto32: According to the function it looks like True... must be from an initial implementation. Will change

reindervdw-cmi (approved):

PR #192, opened 2025-05-05
Mims assembled

frey-perez:


tests/unit/test_mims.py
```diff
@@ -0,0 +1,657 @@
...
+
+from wristpy.core import models
+from wristpy.io.readers import readers
+from wristpy.processing import mims
+
+
+def create_clipped_sine_data(
+    cycles: int = 2, threshold: Union[float, None] = 0.95
+) -> np.ndarray:
+    """Fixture to create dummy sinusoidal data with clipped regions."""
```
Asanto32, line 20: Does this count as a fixture if it doesn't have the @pytest.fixture decorator?

Asanto32:
LGTM just that small comment about the docstring saying it is a fixture ha

PR #194, opened 2025-05-06
Feature/issue 178/build docker image

### LINKED ISSUE #178: Task: Containerize Wristpy
### Description

Make a Docker container that runs wristpy

### Tasks

- [ ] Should take in a actigraphy file/directory and an output directory 

### Freeform Notes

_No response_

Asanto32:
Resolves #178 
Create a docker image of the latest version of wristpy. This PR also include the github action that generates a new Docker container for the latest release (triggered on pushes to main and new version release).

We enable passing environment variables for customization of the docker run.


README.md
```diff
@@ -117,6 +119,59 @@ nonwear_array = subject1.nonwear_epoch
...
+   ```
+
+3. **Run the Docker image** with your data:
+   ```bash
+   docker run -it --rm \
+     -v "/local/path/to/data:/data" \
+     -v "/local/path/to/output:/output" \
+     adamsanto/wristpy
+   ```
+   Replace `/local/path/to/data` with the path to your input data directory and `/local/path/to/output` with where you want results saved.
```
frey-perez, line 139: Leaving this here as a reminder for monday morning, this all looks good to me, but the way the script runs now we can only run directories right? Might be something to consider so that docker image function aligns with the cli/orchestrator.

Asanto32: Actually it should run either since it will simply just call the `poetry run wristpy` with the arguments attached to the ENV_VAR. So it handles it the same way our `orchestrator.run` handles it. 
But maybe I should update the documentation to make that clear.

Asanto32: Actually that might be incorrect now that I test it, because it binds the INPUT_DIR to the internal docker directory `data` and this seems to fail if you try to bind a file to a directory. I'll check if there's a way to handle this by placing the file inside `data` as opposed to mounting the directory

reindervdw-cmi (approved):
LGTM - I'd just make sure you don't include the pngs (and any other large files)

PR #195, opened 2025-05-14
Dropping 3.9 support

Asanto32:


Asanto32:
Removed testing/support for 3.9
Update the pyarrow version since that was limited by 3.9 support (it is the minimum for 3.13 when we make that change)

tests/unit/test_cli.py
```diff
@@ -182,9 +182,8 @@ def test_main_with_bad_epoch(
     )
 
     assert result.exit_code != 0
-    # partial matching due to ANSI escape sequences in Github Actions
-    assert "Invalid value for" in result.output
-    assert "is not in the range x>=1." in result.output
+    assert result.exception is not None
+    assert isinstance(result.exception, (ValueError, TypeError))
```
reindervdw-cmi, line 186: do you need to ensure that both error types are covered? Can it not only be a single one?

Asanto32: Yeah this should only be a TypeError. To be honest I wonder if this will be a short term fix, interested to see what happens if `click` is updated again.

reindervdw-cmi (approved):

Issue #203, opened 2025-05-30
Bug: Checking wristpy version requires an INPUT_PATH


Asanto32:
### Description

Right now, because input is a required argument for typer we cannot check the wristpy version from the terminal without defining an input path. 

### Steps to Reproduce

`wristpy -v` from the CLI

### Expected Results

Should give the version number

### Actual Results

Gives a missing input path error

PR #215, opened 2025-06-09
203 bug checking version

frey-perez:
This addresses issue #203. The changes made were as follows:

1.  The version option was removed from the main function.
2. The --version/ -V option will be handled using typer's app.callback decorator + the is_eager=True argument.  

This allows the version check to happen and exit without running the main function or validating any other options.

### DIFF
--- a/src/wristpy/core/cli.py
+++ b/src/wristpy/core/cli.py
@@ -15,6 +15,22 @@ app = typer.Typer(
 )
 
 
+@app.callback(invoke_without_command=True)
+def version_check(
+    version: bool = typer.Option(
+        False,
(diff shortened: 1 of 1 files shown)


Asanto32 (commented):
Looks good to me, and it seems to be working.

Want to add a quick unit test? Something like this maybe? 

```
def test_main_version(
    create_typer_cli_runner: testing.CliRunner,
) -> None:
    """Test cli version output."""
    result = create_typer_cli_runner.invoke(cli.app, ["--version"])

    assert result.exit_code == 0
    assert "Wristpy version" in result.output
```

Issue #214, opened 2025-06-06
Bug: wristpy-mims reports small numbers when values should be zero.


frey-perez:
### Description

When comparing outputs between wristpy's mims implementation, and the original R implementation values that are reported as 0 in the R version are instead reported as small (but non trivial) positive values. 

### Steps to Reproduce

Run wristpy the mims as the activity metric and compare with the R version output. 

### Expected Results

Generally equal values with very small if any differences. 

### Actual Results

Non-zero values in the R version largely agree with wristpy values. However zero values in the R version are usually small positive values in wristpy.

PR #226, opened 2025-06-27
truncation fix

frey-perez:
This fixes issue #214, a bug where wristpy-mims and the original implimentation of mims seemed to have different cutoff thresholds. 

Previously wristpy had a constant threshold value of 1 * 10^-3, as indicated in the mims paper. Upon closer examination of the original code, it was found that mims actually implements the threshold as the product of the constant value, the sampling rate, and epoch size in seconds, [as shown here](https://github.com/mHealthGroup/MIMSunit/blob/master/R/mims_unit.R#L515).

the wristpy truncation value was changed to match the original implementation.

### DIFF
--- a/src/wristpy/processing/mims.py
+++ b/src/wristpy/processing/mims.py
@@ -685,9 +685,12 @@ def aggregate_mims(
         time=result["time"].cast(pl.Datetime("ns")),
     )
 
+    truncate_threshold = 1e-4 * epoch * sampling_rate
     if truncate:
         aggregated_measure.measurements = np.where(
-            aggregated_measure.measurements <= 0.001, 0, aggregated_measure.measurements
+            aggregated_measure.measurements <= truncate_threshold,
+            0,
+            aggregated_measure.measurements,
         )
 
     return aggregated_measure


src/wristpy/processing/mims.py
```diff
@@ -685,9 +685,12 @@ def aggregate_mims(
         time=result["time"].cast(pl.Datetime("ns")),
     )
 
+    truncate_threshold = 1e-4 * epoch * sampling_rate
```
Asanto32, line 688: Doesn't matter, but being nitpicky I guess technically this should be inside the if statement? It is only used if truncate is true right?

tests/conftest.py
```diff
@@ -68,6 +68,14 @@ def mims_r_version() -> pathlib.Path:
     return pathlib.Path(__file__).parent / "sample_data" / "mims_example_full_run.csv"
 
 
+@pytest.fixture
+def mims_truncated_r_version() -> pathlib.Path:
+    """The actigraphy sample data processed by the mims function in R."""
```
Asanto32, line 73: Data is slightly modified for the truncation test right? Maybe add a small change in docstring to reflect this

Asanto32 (approved):
Looks good to me, just some tiny comments.

Issue #212, opened 2025-06-04
Task: Addressing feedback from collaborative session


Asanto32:
### Description

This is a group of small tasks to improve QoL based on feedback from the collaborative session.

### Tasks

- [ ] Add default output filetype for directory processing.
- [ ] Suppress full traceback for common errors (specifically FileNotFoundError for empty directory)
- [ ] Extend to accept multiple files as input paths (not just one file or one directory)

PR #230, opened 2025-07-02
Feature/Issue 212/Feedback

frey-perez:
This PR addresses issue #212. It addresses the three pieces of feedback we received on quality of life improvements. 

### Default file type when running directories.

Output filetype is now ".csv" by default. All errors regarding an output_filetype being provided when it should have been 'None' have been removed. Behavior is otherwise unchanged. 

### Suppressing full traceback on FileNotFoundError

Added a try/except block to the cli, allowing for the simple printing of the error message when FileNotFoundError is raised.

### Allow for the running of multiple files, instead of entire directories.

Introducing this to the orchestrator was found to add a non-trivial level of complexity to the code base, in order to accomodate what is essentially just calling run_file() in a for loop. As a compromise a snippet of boilerplate code was added to the tutorial for less technical users.

### DIFF
--- a/src/wristpy/core/cli.py
+++ b/src/wristpy/core/cli.py
@@ -74,11 +74,10 @@ def main(
         help="Path where data will be saved. Supports .csv and .parquet formats.",
     ),
     output_filetype: OutputFileType = typer.Option(
-        None,
+        ".csv",
         "-O",
         "--output-filetype",
-        help="Format for save files when processing directories. "
-        "Leave as None when processing single files.",
+        help="Format for save files when processing directories. ",
     ),
     calibrator: Calibrator = typer.Option(
         Calibrator.none,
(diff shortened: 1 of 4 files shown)


src/wristpy/core/cli.py
```diff
@@ -150,17 +149,21 @@ def main(
...
+            output=output,
+            calibrator=calibrator_value,
+            activity_metric=activity_metric.value,
+            thresholds=None if thresholds is None else thresholds,
+            epoch_length=epoch_length,
+            nonwear_algorithm=nonwear_algorithms,  # type: ignore[arg-type] # Covered by NonwearAlgorithm Enum class
+            verbosity=log_level,
+            output_filetype=output_filetype.value,
+        )
+    except FileNotFoundError as e:
```
reindervdw-cmi, line 164: Use a custom error

reindervdw-cmi (changes requested):

reindervdw-cmi (approved):

PR #231, opened 2025-07-23
Update dockerfile

Asanto32:
Change base since buster is no longer supported??!


reindervdw-cmi (changes requested):
How about we add a Github Action to run the unit tests in the Dockerfile -- would prevent this issue and be a good template to have for other repos too.

reindervdw-cmi (approved):

PR #233, opened 2025-07-25
Feature/issue 166/multiple metrics

frey-perez:


src/wristpy/core/orchestrator.py
```diff
@@ -89,19 +100,27 @@ def run(
...
 
-    if activity_metric == "enmo":
-        thresholds = thresholds or (0.0563, 0.1916, 0.6958)
-    elif activity_metric == "mad":
-        thresholds = thresholds or (0.029, 0.338, 0.604)
-    elif activity_metric == "ag_count":
-        thresholds = thresholds or (100, 3000, 5200)
-    elif activity_metric == "mims":
-        thresholds = thresholds or (10.558, 15.047, 19.614)
+    if thresholds is not None:
```
reindervdw-cmi, line 103: Since you do the same check in _run_file, do we need the check here?

src/wristpy/core/cli.py
```diff
@@ -87,22 +117,23 @@ def main(
...
         "--activity-metric",
-        help="Metric used for physical activity categorization. "
-        "Choose from 'enmo', 'mad', 'ag_count', or 'mims'.  ",
+        help="Metric(s) used for physical activity categorization. "
+        "Choose from 'enmo', 'mad', 'ag_count', or 'mims'. "
+        "Use multiple times for multiple metrics: '-a enmo -a mad' etc.",
         case_sensitive=False,
     ),
-    thresholds: tuple[float, float, float] = typer.Option(
+    thresholds: list[str] = typer.Option(
```
reindervdw-cmi, line 129:
```suggestion
    # Typer does not support list[tuple[...]]
    thresholds: list[str] = typer.Option(
```

src/wristpy/core/cli.py
```diff
@@ -62,6 +62,36 @@ def version_check(version: bool) -> None:
...
+    parsed = []
+    for triplet_str in thresholds:
+        parts = triplet_str.strip().split()
+        if len(parts) != 3:
+            raise typer.BadParameter(
+                f"Threshold triplet must have exactly 3 floats: {triplet_str}"
+            )
+        try:
+            values = [float(part) for part in parts]
+            parsed.append((values[0], values[1], values[2]))
```
reindervdw-cmi, line 89: Move this outside the try statement

```suggestion
parsed.append(tuple(values))

```

reindervdw-cmi (approved):
Some minor comments - will pre-approve

src/wristpy/core/orchestrator.py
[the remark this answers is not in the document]
reindervdw-cmi: I just realized there's a slight functional difference between our versions here; mine will treat an empty list `thresholds` as if it were None. Add `is not None` again if you don't want that behavior.

PR #244, opened 2025-09-09
Update dependency, adding permissions

Asanto32:
Fixing some security issues and toml file saying python < 3.13


Asanto32:
@reindervdw-cmi Pretty happy with this now, runs much faster. No longer need the libomp installation for Mac.

I am just using the pre-existing tests though (plug and play compatibility from the previous version). Not sure if you wanted me to break up the detach function and test different components.

.github/workflows/test.yaml
```diff
@@ -9,29 +9,27 @@ on:
...
     - uses: actions/checkout@v4
     - name: Install poetry
       run: pipx install poetry
+
     - uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python_version }}
         cache: pip
 
-    - name: Install dependencies on macOS
```
reindervdw-cmi, line 30: Nice!

pyproject.toml
```diff
@@ -14,17 +14,17 @@ readme = "README.md"
...
 actfast = "^1.1.3"
-numpy = ">=1.26.0, <3.0.0"
+numpy = ">=2.1.0, <3.0.0"
 pydantic = "^2.11.1"
 pydantic-settings = "^2.3.4"
-scipy = "^1.13.0"
-scikit-digital-health = "^0.17.9"
+scipy = "^1.14.1"
 typer = "^0.15.2"
+pandas = "^2.3.2"
```
reindervdw-cmi, line 27: We don't need this explicit dependency anymore, right?

reindervdw-cmi (approved):
Double check you need the pandas (explicit) dependency, but otherwise LGTM

PR #306, opened 2026-06-15
Feature/issue 303/new nonwear sleep interaction

### LINKED ISSUE #303: Task: Handle nonwear/sleep edge
### Description

Currently, it is possible that a SleepWindow can start immediately after a period of NonWear ends. This seems to be unrealistic. It is more likely that this is a false positive for nonwear and the sleepwindow should encapsulate the nonwear.

Need to define a threshold of time where if nonwear and sleep are within that threshold Sleep overwrites the nonwear time and not vice versa. 
 For example, Nonwear period from 1:00 - 1:17, sleep period from 1:05 - 5:00. Current behaviour would modify sleep so that it is now 1:17 - 5:00. However, the wanted behavior would be Sleep from 1:05 - 5:00? or would it be 1:00 - 5:00?  And this nonwear would no longer exist.

The question becomes what is this threshold, for example: Nonwear 12:30 - 12:45, sleep from 12:53 - 5:00? Is that too close? What about nonwear in the middle of a sleep window? The end of a sleepwindow? 

For now make it some parameter passed to the function.

### Tasks

- [ ] Revisit `analytics.sleep_cleanup` function, which currently does the following:
```
    The SleepWindows are first converted to a Measurement object with the same
    timestamps as the reference measurement. Then any overlap with nonwear
    is removed, and finally any blocks of sleep that are less than 15 minutes
    long are removed.
```

### Freeform Notes

_No response_

Asanto32:
Resolves #303 

Begin co-pilot summary:

This pull request introduces significant improvements to the sleep cleanup logic in the analytics module, particularly adding adaptive handling of nonwear periods that immediately precede or are embedded within sleep windows. The changes also update the function signatures and return types, and expand the test suite to ensure robust handling of these new behaviors.

**Enhancements to sleep cleanup logic:**

* The `sleep_cleanup` function in `analytics.py` now supports an adaptive mode that can absorb nonwear periods into sleep if sleep is detected within a configurable buffer after nonwear ends. The function signature and return type have been updated to return both the cleaned sleep and updated nonwear measurements. [[1]](diffhunk://#diff-260a2dc48ffe34c442f4dbdb68f3ea1cffb85a41c8abe1f1d67c75090c0e8407L399-R433) [[2]](diffhunk://#diff-260a2dc48ffe34c442f4dbdb68f3ea1cffb85a41c8abe1f1d67c75090c0e8407L430-R471)

**API and documentation updates:**

* The docstrings for `sleep_cleanup` and `sleep_bouts_cleanup` have been updated to clarify the new behavior, arguments, and return types, reflecting that `sleep_cleanup` now returns a tuple and supports adaptive cleanup, and that `sleep_bouts_cleanup` returns a single measurement. [[1]](diffhunk://#diff-260a2dc48ffe34c442f4dbdb68f3ea1cffb85a41c8abe1f1d67c75090c0e8407L399-R433) [[2]](diffhunk://#diff-260a2dc48ffe34c442f4dbdb68f3ea1cffb85a41c8abe1f1d67c75090c0e8407L439-R495)

**Integration and usage:**

* The orchestrator's `_run_file` method has been updated to unpack the new tuple return value from `sleep_cleanup`, ensuring downstream compatibility.

**Testing improvements:**

* The test suite in `test_analytics.py` has been expanded with several new tests to verify the adaptive sleep cleanup logic, including cases where nonwear is adjacent to sleep, embedded within sleep, separated by a gap, or not adjacent at all. Existing tests have been updated to handle the new tuple return type.

### DIFF
--- a/src/wristpy/processing/analytics.py
+++ b/src/wristpy/processing/analytics.py
@@ -396,8 +396,11 @@ def compute_physical_activty_categories(
 
 
 def sleep_cleanup(
-    sleep_windows: List[SleepWindow], nonwear_measurement: models.Measurement
-) -> models.Measurement:
+    sleep_windows: List[SleepWindow],
+    nonwear_measurement: models.Measurement,
+    adaptive: bool = False,
+    nonwear_sleep_buffer: int = 0,
+) -> Tuple[models.Measurement, models.Measurement]:
     """This function will filter the sleep windows based on the nonwear measurement.
 
     The SleepWindows are first converted to a Measurement object with the same
@@ -405,17 +408,29 @@ def sleep_cleanup(
     is removed, and finally any blocks of sleep that are less than 15 minutes
     long are removed.
 
+    An adaptive second pass can also be performed, where we look ahead a specified
+    number of samples after the end of each nonwear run, and if we find sleep within
+    that buffer, we absorb the nonwear run into sleep. This is to account for some
+    cases where we were finding initial sleep phases to cause false-positives in the
+    non-wear detection.
+
     Args:
-        sleep_windows: List of the sleep windows (Onset/Wake pairs).
+        sleep_windows:  List of the sleep windows (Onset/Wake up pairs).
         nonwear_measurement: The nonwear measurement data used for reference time
             stamps and to remove overlaps with periods of sleep.
+        adaptive: Whether to perform the second pass of absorbing nonwear runs into
+            sleep when sleep immediately follows nonwear. Defaults to False.
+        nonwear_sleep_buffer: The number of samples to look ahead when performing
+            the adaptive sleep cleanup. Defaults to 0.
 
     Returns:
-        A Measurement instance with the cleaned sleep data.
+        A tuple of Measurement instances with the cleaned sleep data and the updated
+        nonwear data.
     """
     logger.debug("Starting the sleep Window cleanup.")
     temporal_resolution = nonwear_measurement.time[1] - nonwear_measurement.time[0]
     num_samples_15min = int(15 * 60 / temporal_resolution.total_seconds())
+    nonwear_arr = nonwear_measurement.measurements.astype(bool)
 
     sleep = _sleep_windows_as_measurement(nonwear_measurement.time, sleep_windows)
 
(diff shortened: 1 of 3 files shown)


Asanto32:
After meeting with Greg; if we use the adaptive pass, nonwear should be modified accordingly (remove the overlap)

src/wristpy/processing/analytics.py
```diff
@@ -396,26 +396,41 @@ def compute_physical_activty_categories(
 
 
 def sleep_cleanup(
-    sleep_windows: List[SleepWindow], nonwear_measurement: models.Measurement
-) -> models.Measurement:
+    sleep_windows: List[SleepWindow],
+    nonwear_measurement: models.Measurement,
+    adaptive: bool = False,
+    nonwear_sleep_buffer: int = 0,
```
reindervdw-cmi, line 402: Can this be optional and None is equivalent to adaptive == False?

src/wristpy/processing/analytics.py
```diff
@@ -396,26 +396,41 @@ def compute_physical_activty_categories(
 
 
 def sleep_cleanup(
-    sleep_windows: List[SleepWindow], nonwear_measurement: models.Measurement
-) -> models.Measurement:
+    sleep_windows: List[SleepWindow],
```
reindervdw-cmi, line 399: Iterable[SleepWindow]

src/wristpy/processing/analytics.py
```diff
@@ -427,7 +442,33 @@ def sleep_cleanup(
         _fill_false_blocks(np.logical_not(filtered_sleep), num_samples_15min)
     )
 
-    return models.Measurement(time=sleep.time, measurements=cleaned_sleep)
+    if adaptive:
```
reindervdw-cmi, line 445: Flip around to "if not adaptive return"

src/wristpy/processing/analytics.py
```diff
@@ -427,7 +442,33 @@ def sleep_cleanup(
...
+        logger.debug(
+            "Sleep Window cleanup first pass complete. Starting adaptive pass."
+        )
+        nonwear_change = np.diff(nonwear_arr.astype(int), prepend=0, append=0)
+        nw_run_starts = np.where(nonwear_change == 1)[0]
+        nw_run_ends = np.where(nonwear_change == -1)[0]
+
+        for s, e in zip(nw_run_starts, nw_run_ends):
```
reindervdw-cmi, line 453: zip(...., strict=True)

tests/unit/test_analytics.py
```diff
@@ -222,5 +222,127 @@ def test_sleep_cleanup() -> None:
...
+        adaptive=True,
+        nonwear_sleep_buffer=0,
+    )
+
+    assert np.all(sleep_result.measurements)
+    assert not np.any(nonwear_result.measurements)
+
+
+def test_new_sleep_cleanup_nonwear_gap_absorbed_only_with_sufficient_buffer() -> None:
+    """Test nonwear + gap is absorbed only when buffer >= gap length."""
```
reindervdw-cmi, line 286: Contradictory statement to the test; either change docstring or change the test

reindervdw-cmi (approved):
Minor comments, but no need for re-review.

PR #317, opened 2026-09-15
EOL 3.10 support (OCT 2026), fixing column inference for csv

### LINKED ISSUE #316: Task: bug fix for column inference of csv
### Description

Reader of csv wasn't properly forcing the schema override to float and was inferring instead, this caused issues when the data was predominantly zeros to start, causing inference to be int instead of float, causing issues later on

### Tasks

- [ ] column names just need to be replaced in `_read_actigraph_csv` as they were actually default values not Acc X/Y/Z

### Freeform Notes

_No response_

Asanto32:
fix column inference bug for actigraph csv

bumping dev and removing 3.10 support since it is EOL oct 2026

Resolves #316


Asanto32:
@frey-perez @reindervdw-cmi 

Remember wristpy? I remember

reindervdw-cmi (approved):
Pepperidge farm remembers.

